PR #159, opened 2013-08-17
HttpResponseForbidden if user is logged in else redirect

yetty:
Hi,
I needed little modification and I wasn't able to do it without modification of ``django-guardian``. I didn't find way how to redirect anonymous users to login page and show ``HttpResponseForbidden`` for users, who are already logged in. So I made this patch.

It adds ``user_specific`` parameter to ``permission_required``. If ``user_specific`` is ``True``, than it will check if user is logged in. If it is so, and user hasn't enough permissions to view the page, error 403 is returned. If user isn't logged in and if anonymous users can't view the page, than he will be redirected to ``login_url``. 

By default, this option is disabled in ``permission_required``. So it keeps backward compatibility. For easier use, I add also ``permission_required_user_specific`` decorator, which is a shortcut to ``permission_required`` with ``user_specific`` parameter turned on.

I don't know, if it fits to django-guardian philosophy, but I posted it here, to save someone else time. 

P.S.: I don't know, if ``user_specific`` is the best name for this functionality. But I didn't find out better label.

### DIFF
--- a/guardian/decorators.py
+++ b/guardian/decorators.py
@@ -98,6 +98,9 @@ def permission_required(perm, lookup_variables=None, **kwargs):
                     lookup_dict[lookup] = kwargs[view_arg]
                 obj = get_object_or_404(model, **lookup_dict)
 
+            if request.user.is_authenticated():
+                return_403 = True
+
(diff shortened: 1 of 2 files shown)


lukaszb:

PR #178, opened 2013-11-26
Add kwarg to get_users_with_perms to filter for specific permissions

lfaraone:
This helps us filter for a list of users who have a particular permission
on an object, not just any permission.

If attach_perms=True is also specified the resultant dictionary value will
still list all the user's permissions, not just the one queried.


lukaszb:

PR #208, opened 2014-04-25
Trying to fix deprecation warnings on Django 1.6.

denilsonsa:
Tries to fix https://github.com/lukaszb/django-guardian/issues/207

Those warnings were introduced in
https://github.com/django/django/commit/6983a1a540a6e6c3bd941fa15ddd8cb49f9ec74e
https://code.djangoproject.com/ticket/15363
https://docs.djangoproject.com/en/1.7/releases/1.6/#get-query-set-and-similar-methods-renamed-to-get-queryset
https://docs.djangoproject.com/en/1.7/releases/1.6/#module-name-model-meta-attribute

Unfortunately, while testing with Python 3.3 and Django 1.6.3, I
couldn't make the tests fail. All 198 tests kept passing both before and
after this commit. I ran the tests usin `python3 setup.py test`.

### DIFF
--- a/benchmarks/run_benchmarks.py
+++ b/benchmarks/run_benchmarks.py
@@ -93,8 +93,9 @@ class Benchmark(object):
         self.objects_count = objects_count
         self.objects_with_perms_count = objects_with_perms_count
 
+        from guardian.compat import get_model_name
         self.Model = model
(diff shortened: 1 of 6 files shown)


lukaszb:
Should be fixed now.

PR #241, opened 2014-07-20
Fixed migrations to allow custom user models

thedrow:
The migrations now allow custom user models to be used with custom (non-integer) primary key fields.
We tested them and they work.
Sorry about the code duplication but I did not find a module where ```get_user_pk_field_fully_qualified_name()``` should be.
Maybe in compat.py?

Please merge and release :)

[Review on Reviewable](https://reviewable.io/reviews/lukaszb/django-guardian/241)

### DIFF
--- a/guardian/migrations/0001_initial.py
+++ b/guardian/migrations/0001_initial.py
@@ -7,6 +7,13 @@ from guardian.compat import get_user_model
 
 User = get_user_model()
 
+def get_user_pk_field_fully_qualified_name():
+    pk_field = User._meta.pk
+    module = pk_field.__class__.__module__
+    if module is None:
+        return pk_field.__class__.__name__
+    return module + '.' + pk_field.__class__.__name__
+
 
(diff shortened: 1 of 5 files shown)


brianmay:
This pull request has significant changes to south migrations, which are depreciated. If it is still relevant, needs to be redone. This pull request should be closed.

thedrow:
It is if we still support Django 1.6 or projects that upgraded from 1.1.6 and still use south migrations.
What needs to be redone?

brianmay:
There is a ticket proposing to drop support for Django < 1.7 - #350.

Django < 1.7 is official end-of-life - see https://www.djangoproject.com/download/ - and probably has security issues.

It might seem harsh, however I don't think we can keep supporting all old Django versions especially after they stop being supported upstream. It adds complexity to the code base, is likely to break without notice, and takes developer time away from improving django-guardian further.

PR #246, opened 2014-07-27
Allow specifying permission for shortcuts.get_users_with_perms

mitar:
This pull requests add a `permission_name` to `get_users_with_perms` so that one can get all users which have a given permission instead of any permission.


lukaszb:

Moreover, I'm not that sure of this one anymore. It would probably make more sense to allow to narrow permissions to a set of codenames instead of only one.

mitar:
Sorry, didn't have time to look into this.

PR #247, opened 2014-07-27
Allow specifying an empty list of permissions for get_objects_for_user

### LINKED ISSUE #244: get_objects_for_user should allow get objects for any permission
One would assume that you can run `get_objects_for_user(user, [], klass=Model, any_perm=True)` and get all objects of `Model` for which an user have any permission over. Currently it raises an exception.

In this way it will also match `get_users_with_perms`, which returns users with any permission for an object.

mitar:
This allows one to ask for all objects for which an user has any permission. In a way it mirrors `get_users_with_perms`.

In addition, I fixed how the final query is made. Instead of creating an in-memory list of all IDs and then passing it in, you can pass all IDs simply as a subquery. This both improves the query and also allows more complicated basic queries (I was getting an error that I cannot use OR for two queries with "extra"). Also, it works now with documents which do not have integers as primary keys (I have UUIDs and `int(v)` does not work).

Fixes #244.


gabn88:
Why is this never pushed into django-guardian?

brianmay:
As far as I can tell, this pull request makes a number of changes within the one commit. This makes it very hard to review. The comment suggests it is adding support for perms to be an empty list, however I can't see anything that is related to this. Rather it seems to change the behavior of `ctypes` and `codenames` and I am failing trying to understand this.

Test cases for this would also be good.

mitar:
Yes, it was hard to make this separate things. As I explained, there are three things here:
* allowing to get objects for which user has any permission
* passing query as a sub-query instead of memory list
* allowing non-numeric IDs

mitar:
And no, it does not change behavior of `ctypes` and `codenames`, it just makes it a subquery instead of a list.

guardian/shortcuts.py
```diff
@@ -393,8 +396,9 @@ def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=Fals
     user_model = get_user_obj_perms_model(queryset.model)
     user_obj_perms_queryset = (user_model.objects
         .filter(user=user)
-        .filter(permission__content_type=ctype)
-        .filter(permission__codename__in=codenames))
+        .filter(permission__content_type=ctype))
+    if len(codenames):
+        user_obj_perms_queryset = user_obj_perms_queryset.filter(permission__codename__in=codenames)
```
mitar: See here. Instead of just making it always filter by `codenames`, it filters only when it is non-zero (so  when `ctype` was specified). This allows the behavior of getting any objects for any permission.

guardian/shortcuts.py
```diff
@@ -404,15 +408,18 @@ def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=Fals
         group_model = get_group_obj_perms_model(queryset.model)
         group_filters = {
             'permission__content_type': ctype,
-            'permission__codename__in': codenames,
             'group__%s' % get_user_model().groups.field.related_query_name(): user,
         }
+        if len(codenames):
+            group_filters.update({
+                'permission__codename__in': codenames,
+            })
```
mitar: Similar here. Just splitting filter building into two steps, where one is conditional.

guardian/shortcuts.py
```diff
@@ -430,17 +437,11 @@ def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=Fals
         counts = user_obj_perms_queryset.values(fields[0]).annotate(object_pk_count=Count(fields[0]))
         user_obj_perms_queryset = counts.filter(object_pk_count__gte=len(codenames))
 
-    values = user_obj_perms_queryset.values_list(fields[0], flat=True)
-    if user_model.objects.is_generic():
-        values = [int(v) for v in values]
-    objects = queryset.filter(pk__in=values)
+    q = Q(pk__in=user_obj_perms_queryset.values(fields[0]))
```
mitar: And this one is just making it a subquery instead of in-memory list.

mitar: And non-integer IDs you get for free.

mitar:
I added comments with more explanation.

brianmay:
Thanks for the comments. Will look at them now.

brianmay:
@mitar Can I please have your opinion on #334? Suspect it might also be a patch to deal with perms being an empty list too. Are both pull requests required? Just this one?

This change seems OK to me. It needs to be fixed so that it will apply cleanly, and it really should have test cases too.

mitar:
I updated the pull request. Let's me see about the tests.

mitar:
The pull request here does not fix the #334. The test available in #334 still fails. So that is something else.

mitar:
I added a test.

mitar:
Tests for non-numeric IDs were already there. Passing query as a sub-query instead of memory list is just an optimization, but no other tests broke, so it seems it is a good optimization. :-)

brianmay:
Thanks for this. Looks good to me.

I don't have write access yet, hope to get it very soon. Unless @lukaszb raises any objections I will rebase this - to avoid the merge commit inside the pull request which from my experience is a bad idea - and merge the result.

brianmay:
I rebased it, compared it with your original, and created a new merge request #363. That makes this merge request redundant now. Unless there are any problems getting #363 merged.

mitar:
Great! Thanks!

brianmay:
Merged #363, so closing this.

mitar:
Great!

PR #264, opened 2014-09-16
Commented out USERNAME_FIELD

thanhleviet:
This caused error: TypeError: create_superuser() takes exactly 4 arguments (3 given)
I removed this line (following the solution here http://stackoverflow.com/questions/17342702/cant-create-superuser-with-custom-userprofile), it worked


lukaszb:
Looks like there are more problems with new Django version and example project.

This change in particular doesn't look to solve a problem anyway. Closing as we need more "proper" solution.

PR #278, opened 2014-11-05
Update assign.rst - syncdb is not enough

chiller:
source: http://stackoverflow.com/questions/19277543/working-with-django-model-permission-raises-doesnotexist-error
new permissions are not created in Django 1.4.10 with default syncdb


lukaszb:
`--all` is south related only switch - we assume that south users would know what they are doing. However, I do agree that we should also try to educate users if possible so I've extended that doc. Thanks for noticing this!

PR #295, opened 2015-01-27
Cached the object permissions models in order to save queries

thedrow:
I noticed that guardian performs extra queries when they are not needed.
Caching them in process is probably a good idea.

[Review on Reviewable](https://reviewable.io/reviews/lukaszb/django-guardian/295)


thedrow:
How the hell this failure happened? https://travis-ci.org/lukaszb/django-guardian/jobs/48511254#L200
I'm patching at the function level.

brianmay:
Caching is probably a good thing. It is also very good that this pull requests has tests to go with it too.

However, it is a rather complicated change, and I suspect there are more urgent things that need fixing first. So suggest we look at the more urgent things and revisit this one later.

PR #297, opened 2015-02-14
Django migrations south+1.7

patgmiller:
Django 1.7 wouldn't do any migrations. I wasn't getting any errors, it just it wouldn't run any migrations for Guardian. So rather than always doing it by hand, this will make it simpler. 

As per Django / South documentation South will look in `south_migrations` before migrations. I believe it requires South>=1.0 for this. 

[Migrations - Library / Third-party Apps](https://docs.djangoproject.com/en/1.7/topics/migrations/#libraries-third-party-apps)


lukaszb:
Makes sense, thank you!

Issue #307, opened 2015-04-16
Django 1.8 GuardedModelAdminMixin.get_queryset exception


lalstef:
Hi,

I'm using django-guardian (through django-userena) with Django 1.8. When trying to access User model in django admin, GuardedModelAdminMixin raises the following Exception:

`'super' object has no attribute 'queryset'`

In `GuardedModelAdminMixin.get_queryset` the problematic code is:
```
    def get_queryset(self, request):
        # Prefer the Django >= 1.6 interface but maintain
        # backward compatibility
        method = getattr(
            super(GuardedModelAdminMixin, self), 'get_queryset',
            super(GuardedModelAdminMixin, self).queryset)

```
In Django 1.8 (I guess in all Django >= 1.6) `queryset` method does not exist, but this part

`super(GuardedModelAdminMixin, self).queryset` is evaluated and raising Exception even though `get_queryset` method exists and it's found.

frwickst:
The problem here is that `super(GuardedModelAdminMixin, self).queryset` is not defined in Django 1.8. Even if get_queryset is, getattr will still fail as we have a undefined variable as the last paramter.
Example of what the code basically does:
```python
class Test(object):
    a = 'test string'

t = Test()
getattr(t, 'a')
>> 'test string'
getattr(t, 'a', Test.b)
>> AttributeError: type object 'Test' has no attribute 'b'
```
What we could do is something like:
```
    def get_queryset(self, request):
        # Prefer the Django >= 1.6 interface but maintain
        # backward compatibility
        method = getattr(
            super(GuardedModelAdminMixin, self), 'get_queryset',
            getattr(super(GuardedModelAdminMixin, self), 'queryset', None))
    qs = method(request)
```
This of course can make the code that calls `method` crash as well, as if we set it to None the `qs` declaration would fail. This might not we a big issue however as `method` should not be set to None in any known Django version to date.
I'll create a pull request with this fix and see if @lukaszb accepts this solution.

mcgeeco:
Would be great to get this merged. Just out of curiosity - is there really a need to maintain compatibility with such old versions of Django on new releases? Especially since it's stuff Django itself has removed back-compatibility for.

HEdingfield:
@mcgeeco Agreed, seems like the docs could just point users of 1.6 or whatever to download an older version.

Karlheinzniebuhr:
having the same issue, solution didn't work for me

autrilla:
[...]

PR #311, opened 2015-04-21
Use getattr to check is GuardedModelAdminMixin has the attribute queryset

frwickst:
We can not check for queryset in Django 1.8 as the attribute is no longer available
The getattr function still checks the default parameter and fails with a `AttributeError`
as `super(GuardedModelAdminMixin, self)` has no attribute `getquery`. We also need to
pass a default parameter (`None`) to the second getattr call as this will otherwise fail
as well.

This fixes: #307

### DIFF
--- a/guardian/admin.py
+++ b/guardian/admin.py
@@ -66,7 +66,7 @@ class GuardedModelAdminMixin(object):
         # backward compatibility
         method = getattr(
             super(GuardedModelAdminMixin, self), 'get_queryset',
-            super(GuardedModelAdminMixin, self).queryset)
+            getattr(super(GuardedModelAdminMixin, self), 'queryset', None))
         qs = method(request)
 
         if request.user.is_superuser:


ayc92:
I think you also need to remove the `django.VERSION < 1.6` check, otherwise in line 156 of `obj_perms_manage_view`, `self.queryset` will error out.

frwickst:
The above commits are just a rebase to the current 'devel' branch, no changes to the code yet.

frwickst:
Would it be a better solution to change the queryset call at line 151 (and 224,227) to get_queryset instead? Ideas?

ayc92:
@frwickst I'm in favor of that solution.

frwickst:
Ad pointed out in Issue #307 this might be a good time to remove support for older versions of Django. But for this to happen I want the go ahead from @lukaszb before going forward.

TauPan:
As a user: I'm also in favour of changing self.queryset in lines 151, 224 and 227 to self.get_queryset ... I don't see how these changes would hurt backwards compatibility, as the calls are specific to GuardedModelAdminMixin and decendants, and django < 1.6 apps will have the aliases. I'm going to deploy django-guardian in my Django 1.8 installation that way and notify you if I see any further problems.

evandavey:
With 1.8 and on the guardian devel branch as at https://github.com/lukaszb/django-guardian/commit/112c373f213a19d93baa81fa4a941a41333115b5, I needed to change all lines guardian/admin.py
```
obj = get_object_or_404(self.queryset(request), pk=object_pk)
```
to
```
obj = get_object_or_404(self.get_queryset(request), pk=object_pk)
```

Issue #309, opened 2015-04-17
Django 1.9 Deprecation Warnings


cyberjacob:
Django-guardian is generating the following warnings in Django 1.8:

````
/usr/local/lib/python2.7/dist-packages/guardian/models.py:8: RemovedInDjango19Warning: django.contrib.contenttypes.generic is deprecated and will be removed in Django 1.9. Its contents have been moved to the fields, forms, and admin submodules of django.contrib.contenttypes.
  from django.contrib.contenttypes.generic import GenericForeignKey

/usr/local/lib/python2.7/dist-packages/guardian/compat.py:7: RemovedInDjango19Warning: django.utils.importlib will be removed in Django 1.9.
  from django.utils.importlib import import_module
````

Should be able to fix these by changing the imports to `from django.contrib.contenttypes.fields import GenericForeignKey` and `from importlib import import_module` respectively.

brianmay:
Looks like the related pull request was merged, so closing.

PR #312, opened 2015-04-22
Use importlib and import GenericForeignKey from from django.contrib.contenttypes.fields

frwickst:
Importlib is included in Python 2.7, and will be dropped in Django 1.9 as it is no longer needed

django.contrib.contenttypes.generic will be removed in Django 1.9 and has been replaced by
fields, forms, and admin submodules instead of generic.

This should hopefully fix: #309
Kinda hard to test when all the Django 1.8 stuff ain't working yet. I'll fix it if there is a need to after the Django 1.8 pull-requests are merged :)

### DIFF
--- a/guardian/compat.py
+++ b/guardian/compat.py
@@ -4,9 +4,13 @@ from django.conf import settings
 from django.contrib.auth.models import Group
 from django.contrib.auth.models import Permission
 from django.contrib.auth.models import AnonymousUser
-from django.utils.importlib import import_module
 import six
 import sys
+
+try:
+    from importlib import import_module
+except ImportError:
+    from django.utils.importlib import import_module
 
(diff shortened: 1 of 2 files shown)


TauPan:
I just verified that the deprecation warnings I got with django-admin check are fixed with this PR.

lukaszb:
Thanks @frwickst and @TauPan !

xordoquy:
Would it be possible to get a release that includes this fix ?
Would be much appreciated to go forward with Django REST Framework against master.

lukaszb:
Yup, just started vacations, hopefully I would release new version soon.

xordoquy:
Looking forward to it and enjoy your vacations !

PR #318, opened 2015-05-21
global perms carry over in has_perm, fixes #294

### LINKED ISSUE #294: accept_global_perms for has_perm
Hi,
I think it's a good idea to add something like accept_global_perms in the permission_required decorator to has_perm as well. Neither django default authentication backend nor guardian's returns True when user has global perm and an object is given.

ryuusenshi:
This PR adds a new setting: `GUARDIAN_GLOBAL_PERMISSIONS_CARRY_OVER`. The default is set to `False` to maintain backwards compatibility.

If this setting is set to `True`, it makes `has_perm` return `True` if an object check is performed and a global perm exists (regardless of whether the object perm exists or not).

``` python
assign_perm('auth.change_group', user)
user.has_perm('change_group', group)
# returns True even without assign_perm('change_group', user, group)
```

In my opinion, this behavior should be controlled by a project setting and not a `has_perm` parameter, since the expectation for this kind of behavior is usually defined on the level of the project and should be immutable.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -54,3 +54,4 @@ Authors ordered by first contribution
 - Verena Jaspersen <verena.jaspersen@gmail.com>
 - Bertrand Svetchine <bertrand.svetchine@gmail.com>
 - Frank Wickström <frank@bambuser.com>
+- David Filipovic <david.filipovic@gmail.com>
--- a/guardian/conf/settings.py
+++ b/guardian/conf/settings.py
@@ -17,6 +17,8 @@ TEMPLATE_403 = getattr(settings, 'GUARDIAN_TEMPLATE_403', '403.html')
 RAISE_403 = getattr(settings, 'GUARDIAN_RAISE_403', False)
 GET_INIT_ANONYMOUS_USER = getattr(settings, 'GUARDIAN_GET_INIT_ANONYMOUS_USER',
     'guardian.management.get_init_anonymous_user')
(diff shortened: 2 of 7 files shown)


brianmay:
This references #294.

brianmay:
Good to see this pull request has tests. Unfortunately, this looks like it includes tests that look like they should be applied independently... Such as GroupHasPermTest and PermQualifiedNameTest don't look related to this pull request.

What is the use case for global permissions? Not sure I understand why this is a desirable feature.

ad-m:
Global permissions in general is usable for example to create moderators which have permissions to change all posts (global permission to ```change_post```), but unable to delete posts (in opposite of superuser) and can change only posts. At the same time users can have granted privilege to change own post.

Natgho:
Wow, 10 years of PR, huh? :)
@ryuusenshi or @kmmbvnr If you revise this PR according to the latest version, I'm eager to review it.

thclark:
I think this PR should probably be closed but used as inspiration in a larger piece of work looking at the mass of confusion around global versus object permissions - there's a mass of issues up.

Natgho:
Actually, the logic behind this PR is quite good. Synthesizing global permissions with django-guardian permissions could be a nice feature. We can include this feature in our feature list, but it needs to be improved, taking into account potential issues such as naming conflicts. 

If @ryuusenshi  takes the time to update it and synchronize it with the current code base, we can include it in the next release.

Natgho:
https://github.com/django-guardian/django-guardian/issues/1025

PR #319, opened 2015-05-27
Django 1.8 compatibility: Direct foreign key (performance) optimizati…

qdotme:
…on support and compatibility

The current Django 1.8 compatibility does not support the case where 
http://django-guardian.readthedocs.org/en/latest/userguide/performance.html features are employed - this code changes the condition in the utility function used for determining which model is the permission store.

### DIFF
--- a/guardian/utils.py
+++ b/guardian/utils.py
@@ -26,6 +26,8 @@ from guardian.exceptions import NotUserNorGroup
 
 from django.contrib.auth.views import redirect_to_login
 
+import django
+
 logger = logging.getLogger(__name__)
 abspath = lambda *p: os.path.abspath(os.path.join(*p))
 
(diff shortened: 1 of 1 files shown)


qdotme:
We could also think about a test case for this - unfortunately I'm not too familiar with the django-guardian test setup.

soby:
+1, just ran into this. It's returning the generic classes instead of the custom classes

7wonders:
Same here. Generics working well but no direct fk.

k4nar:
I just ran into this issue, and this patch fixes it.

Ghabry:
Just wasted two hours to figure out why my permission checks don't work anymore after updating to 1.8.
Same issue here :/

gabn88:
When is this going to be pulled into the release?

brianmay:
Looks like I got this applied in #359, see bac7c403062e81594a2ebed31562e72e83004dff

This pull request can be closed.

PR #321, opened 2015-06-03
Change self.queryset(request) to self.get_queryset(request) in admin

frwickst:
I seemed to have missed this in #311. Hopefully you still have a chance to merge this before releasing a new version. 

People have noted that these changes needs to be made for the admin.py file to be properly fixed. This will however break compatibility with some older versions of Django. Lets see what Travis has to say...


brianmay:
I believe this pull request is no longer required and can be closed.

PR #322, opened 2015-06-04
Fixed bug with app_label==None

giocalitri:
I'm not sure why this is a problem, but I encountered this problem in some unit-tests for my app.
While I was reading your unit-tests for guardian.shortcuts.get_objects_for_user I realized that maybe this situation (app_label == None) should never occur.
If this is the case, maybe you need to raise an exception before, but I'm not sure of the right logic here.

### DIFF
--- a/guardian/shortcuts.py
+++ b/guardian/shortcuts.py
@@ -413,7 +413,7 @@ def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=Fals
     # a superuser has by default assigned global perms for any
     if accept_global_perms and with_superuser:
         for code in codenames:
-            if user.has_perm(app_label + '.' + code):
+            if app_label is not None and user.has_perm(app_label + '.' + code):
                 global_perms.add(code)
         for code in global_perms:
             codenames.remove(code)


guardian/shortcuts.py
```diff
@@ -413,7 +413,7 @@ def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=Fals
     # a superuser has by default assigned global perms for any
     if accept_global_perms and with_superuser:
         for code in codenames:
-            if user.has_perm(app_label + '.' + code):
+            if app_label is not None and user.has_perm(app_label + '.' + code):
```
damycra: `if user.has_perm(ctype.app_label + '.' + code):`  # better as app_label will always be None if passing in simple codenames (e.g. 'change_group') along with `klass` parameter

brianmay:
`app_label` could be None at this point if `perms` is an empty list. So is having `perms` an empty list valid?

This might be related to  #247 - "Allow specifying an empty list of permissions for get_objects_for_user" a pull request that I didn't understand.

Not sure I fully understand just yet what is going on in this code.

brianmay:
Suggest this should be closed in favor of #334, which also has a test case.

Issue #327, opened 2015-06-09
get_users_with_perms(obj, attach_perms=True, with_superusers=False) seems inconsistent


kewama:
Hi,

Using django-guardian 1.2.5.

When a superuser has no specific permission for a particular object, the get_users_with_perms shortcut works as expected:

```python
>>> from guardian.shortcuts import *
>>> from vrt.testnet.models import Testnet
>>> from django.contrib.auth import get_user_model
>>> u = get_user_model().objects.get(username='aschultz')
>>> u.is_superuser
True
>>> t = Testnet.objects.get(pk=4)
>>> get_users_with_perms(t, attach_perms=True, with_superusers=False, with_group_users=False)
{}
```

But if we grant that user a particular permission to that object, get_users_with_perms then reports that the user has *all* permissions for it:

```python
>>> assign_perm('access_testnet', u, t)
<UserObjectPermission: snapshot-4 | aschultz | access_testnet>
>>> get_users_with_perms(t, attach_perms=True, with_superusers=False, with_group_users=False)
{<User: aschultz>: [u'access_testnet', u'add_testnet', u'change_testnet', u'create_snapshot', u'delete_testnet', u'manage_testnet']}
```

This seems a surprising behavior. I would have expected this instead:

```python
>>> get_users_with_perms(t, attach_perms=True, with_superusers=False, with_group_users=False)
{<User: aschultz>: [u'access_testnet']}
```

Thanks,
Kevin

brianmay:
There seems to be a bit of confusion in the code regarding the global permissions. When we get the user list we only check for object permissions. Then when we retrieve the permissions we merge the global permissions and the object permissions.

We need to be a bit more consistent as to if we use the global permissions or not.

brianmay:
Have added the 'API change' label to this, as fixing this is going to change the API which people may be relying on. So fixing should perhaps be done in a major release.

brianmay:
Moving to #380.

Issue #46, opened 2011-08-14
get_users_with_perms - with_group_users = False and attach_perms = True returns group permissions.


hackoder:
Calling get_users_with_perms with with_group_users = False and attach_perms = True still returns those permissions that the user has through a group.

Test case:
- Create a user and grant 'read' permissions to an object
- Assign user to group.
- Assign group 'write' permissions to object.

get_users_with_perms(with_group_users = False, attach_perms = True) returns both 'read' and 'write'.

mitar:
I think this was fixed with #403.

kewama:
It is fixed in version 1.4.2.

PR #328, opened 2015-06-10
Add support for "direct_perms_only" option

kewama:
Add support for "direct_perms_only" option

This option identifies the permissions which are *directly* assigned
to a user for an object. It disregards permissions that the user may
have acquired via group memberships or superuser status.

The new option is disabled by default for backwards compatibility.

Two new unit tests are added to demonstrate behavior with and without
the option.

The purpose of the option is to allow initialization of permission
forms used to grant specific permissions to specific users,
independently of their superuser or group memberships. In other words,
I need to be able to tell if an individual user has been specifically
granted a permission so I can show this on the form, regardless of
whether the user *also* has this permission via group membership or
superuser status.

This feature responds to a need expressed in issues #327 and #46.

### DIFF
--- a/guardian/shortcuts.py
+++ b/guardian/shortcuts.py
@@ -137,12 +137,23 @@ def remove_perm(perm, user_or_group=None, obj=None):
         model = get_group_obj_perms_model(obj)
         model.objects.remove_perm(perm, group, obj)
 
-def get_perms(user_or_group, obj):
+def get_perms(user_or_group, obj, direct_perms_only=False):
     """
     Returns permissions for given user/group and object pair, as list of
     strings.
+
+    :param user_or_group: instance of ``User``, ``AnonymousUser`` or ``Group``;
+      passing any other object would raise
+      ``guardian.exceptions.NotUserNorGroup`` exception
+
+    :param obj: persisted Django's ``Model`` instance
+
+    :param direct_perms_only: If set to ``True`` and ``user_or_group`` is a
+       ``User`` instance, result would contain only permissions assigned
+       directly to the user for the given ``obj``, not those coming via user's
+       superuser status or group memberships.
     """
-    check = ObjectPermissionChecker(user_or_group)
+    check = ObjectPermissionChecker(user_or_group, direct_perms_only)
     return check.get_perms(obj)
 
 def get_perms_for_model(cls):
@@ -159,7 +170,7 @@ def get_perms_for_model(cls):
     return Permission.objects.filter(content_type=ctype)
 
 def get_users_with_perms(obj, attach_perms=False, with_superusers=False,
-        with_group_users=True):
+        with_group_users=True, direct_perms_only=False):
     """
     Returns queryset of all ``User`` objects with *any* object permissions for
     the given ``obj``.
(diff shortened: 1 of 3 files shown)


brianmay:
Seems fine to me - apart from title which I find confusing. Good tests. However, adds 71 lines - is it worth the added complexity?

Also:
* #46 - has already been closed as invalid.
* #327 - I can't see the connection.

kewama:
Brian, thanks for fixing my title.

Using direct_perms_only=True provides a way to determine the permissions directly assigned to a user, vs. permissions available to that user due to superuser status (#327) or group membership (#46). This feature is essential to my site.

mitar:
I do not think this is needed anymore with #403 being merged in.

guardian/shortcuts.py
```diff
@@ -159,7 +170,7 @@ def get_perms_for_model(cls):
     return Permission.objects.filter(content_type=ctype)
 
 def get_users_with_perms(obj, attach_perms=False, with_superusers=False,
-        with_group_users=True):
+        with_group_users=True, direct_perms_only=False):
```
mitar: Isn't `direct_perms_only` redundant? You could simply set `with_superusers` and `with_group_users` to `False` and you would get the same result as setting `direct_perms_only` to `True`? No?

mitar: Which is what you do below. :-)

brianmay:
Going to close this based on comments on @mitar. If this is incorrect and #403 does not solve the required problem, please let me know and I will reopen.l

guardian/shortcuts.py
kewama: #403, #407 do solve my issue. I think the new get_users_with_perms() behavior is great and what I originally expected (see #327). However, it is a backwards-incompatible change and so may break existing apps that depend on the previous behavior. Probably worth calling out in the release notes.

 Arguably, the new behavior could be considered a bug fix, but there may be applications depending on the new behavior that will break when upgraded.

mitar: In which way it is a backwards incompatible change? It didn't break any tests.

You mean, because now `attach_perms` set to `True` return really what you asked for and not everything as before? I see this as a bugfix, but yes, it should be mentioned in the release notes.

kewama: There may not have been tests that cover it, but the old behavior was described in #46, which was closed as invalid. Thus a guardian user might have felt comfortable depending on that behavior. Or  might be accidentally depending on it and be surprised on upgrade.

The defense of the old behavior in #46 was the reason I created a pull request that did not break backwards compatibility. But treating it as a bugfix results in a better long term solution.

mitar: Good point. I opened: https://github.com/django-guardian/django-guardian/pull/410

brianmay: Am beginning to think we need a -devel mailing list hosted somewhere so we can discuss general strategies involving django-guardian. What constitutes an backward incompatible change, when new releases should be made, should we have a section on the documentation (like Django) that describes all potential compatibility issues, etc. Any good mailing list providers for this purpose? I only know the big ones like Google and Yahoo. alioth.debian.org might also be a good option.

mitar: Personally, I prefer to keep all stuff/discussions in one place (GitHub issues/pull requests). Moreover, I observed that younger developers have issues using mailing lists. They are simply old school.

So I do not see much benefit of having a mailing list.

Maybe we just need label "discussion" for an issue and then we discuss something there.

PR #340, opened 2015-08-10
Added new model GuardianGroup which subclasses django Group to enable has_perm check for Group objects

shinjanxp:


brianmay:
This change really could do with test cases to go with it.

brianmay:
I don't think this change has been thought out or tested. By inheriting from Group, you are using Multi-table inheritance, and creating a new table. See https://docs.djangoproject.com/en/1.8/topics/db/models/#multi-table-inheritance

Suspect what you may have wanted instead was a proxy model, see https://docs.djangoproject.com/en/1.8/topics/db/models/#proxy-models

Sure, this is easy to do. However no point adding complexity for no reason. What is the use case for a group.has_perm() function?

shinjanxp:
Yes. I suppose a proxy model would have been apt for this case. I had implemented this function in a side project where there was a 2-level permission checking system. User - Group - Model. This required a check whether the group had a particular permission on the model, and whether a user had a certain level of access with respect to the group.

brianmay:
I am going to close this for now. Feel free to resubmit using a Proxy model :-)

k0t3n:
@brianmay is it already possible to inherit permissions from the group or has this PR remained unresolved?

brianmay:
I am no longer maintaining this package. Looking at what I wrote above it seems I had an alternative suggestion that would have solved the reported use case. If you have a different use case, please open another ticket.

brianmay:
Sorry wrote previous text (now deleted) before looking at the pull request.

I am no longer maintaining this package.

As far as I know, nothing happened here, because I was unhappy with the quality of the pull request, and so far nobody has updated the pull request as per my feedback to address the issues.

This pull request is about adding a has_perm to groups.

When you say "inherit permissions" I suspect you might want something different entirely, that was never provided for by this pull request, in which case you should open a new issue.

PR #339, opened 2015-08-09
guardian | templatetags - guardian_tags fix compability in Django 1.9

s0ng:

PR #353, opened 2015-10-25
Django 1.9 compatability

aidanlister:
I wasn't able to find out where get_template was removed, but this commit cleans up a few places where it was removed:
https://github.com/django/django/commit/24dffaf0cbe416167f99afd1d47f20732929f758

Either way this helper seems like it could live in it's own file, it does not seem like a mainstream helper.

### DIFF
--- a/guardian/templatetags/guardian_tags.py
+++ b/guardian/templatetags/guardian_tags.py
@@ -8,14 +8,6 @@
 from __future__ import unicode_literals
 from django import template
 from django.contrib.auth.models import Group, AnonymousUser
-try:
-    # Django < 1.8
-    from django.template import get_library
-    from django.template import InvalidTemplateLibrary
-except ImportError:
-    # Django >= 1.8
-    from django.template.base import get_library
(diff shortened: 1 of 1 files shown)


lukaszb:
@aidanlister hey, thanks for the input however changes break compat with other versions...

In addition, please add 1.9 check to travis/tox configurations.

brianmay:
The Django 1.8 tests weren't actually executed by travis-ci. "Ran 0 tests in 0.000s". If they were executed they would probably be fails too. See #356.

brianmay:
It looks like this change unconditionally removes the `friendly_load` tag. However a number of files reference `friendly_load`, so the change is invalid.

I don't like the idea of `friendly_load`, it makes me think of conditions were the file should exist but disappears and the error gets silently ignored, however removing it isn't as simple as this pull request makes out.

brianmay:
Suggest we close this request. I have not yet looked at Django 1.9 compatibility, however this doesn't look like the way to do it.

brianmay:
See #339, looks like it might be a better solution.

PR #354, opened 2015-10-30
Object IDs with special characters

Freebird2014:
Object IDs with special characters，
 Underscores in primary key get quoted to "_5F"


Freebird2014:
my object ids is pd_ios, and i get the url of the item is pd_5Fios, and this causes a 404 every time I  want to manage this object permission.

brianmay:
Think this pull request should get merged. Seems strange it is required, however that is due to Django admin's URL scheme.

brianmay:
Merged in af3b1f3a7f21bfe484f9c33581d4b77907072413.

PR #361, opened 2015-11-08
Use post_migrate if possible, otherwise post_syncdb.

audreyr:
In Django 1.9, `post_syncdb` is gone. The replacement is `post_migrate`.

This PR addresses it by:
* Trying to use `post_migrate` if possible.
* Keeping `post_syncdb` as a fallback, so that this still works on Django 1.6 and earlier

### DIFF
--- a/guardian/management/__init__.py
+++ b/guardian/management/__init__.py
@@ -43,5 +43,11 @@ def create_anonymous_user(sender, **kwargs):
 
 # Only create an anonymous user if support is enabled.
 if guardian_settings.ANONYMOUS_USER_ID is not None:
-    signals.post_syncdb.connect(create_anonymous_user, sender=guardian_app,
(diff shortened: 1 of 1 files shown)


brianmay:
Why was this merge request closed?

brianmay:
Ok, I see, looks like it causes the tests to fail under Django 1.7 and Django 1.8.

Regardless the problems this pull requests attempts to fix are real, so am going to keep this open as a reminder that this needs fixing.

brianmay:
Looks like post_migrate isn't getting invoked as expected before the tests.

brianmay:
The sender argument was incorrect, needs to be a a Django 1.7 style App object. Fixed and merged in cf1c818d30abaa649eae2e5697e862c168d8b166

PR #363, opened 2015-11-12
Allow specifying an empty list of permissions for get_objects_for_user

brianmay:
Rebased from #247.


brianmay:
This change is causing problems with postgresql, see #366.

Issue #364, opened 2015-11-14
example_project broken


brianmay:
This is used by some of the tox tests. It is not used by the travis tests.

```
CommandError: One or more models did not validate:
testapp.customuser: Accessor for m2m field 'groups' clashes with related m2m field 'Group.customuser_set'. Add a related_name argument to the definition for 'groups'.
testapp.customuser: Accessor for m2m field 'user_permissions' clashes with related m2m field 'Permission.customuser_set'. Add a related_name argument to the definition for 'user_permissions'.
core.customuser: Accessor for m2m field 'groups' clashes with related m2m field 'Group.customuser_set'. Add a related_name argument to the definition for 'groups'.
core.customuser: Accessor for m2m field 'user_permissions' clashes with related m2m field 'Permission.customuser_set'. Add a related_name argument to the definition for 'user_permissions'.
```

Wtower:
To elaborate on this, I tried to evaluate for different versions. Regarding Python, results are consistent for both 2.7 and 3.4.

With Django 1.5 and 1.6 the error is as @brianmay describes above, and this also occurs on `syncdb` and not only tests.

With Django 1.7 and 1.8 we get:

    Traceback (most recent call last):
      File "./manage.py", line 17, in <module>
        execute_from_command_line(sys.argv)
      File "/home/gkarak/.virtualenvs/django-guardian3/lib/python3.4/site-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
        utility.execute()
      File "/home/gkarak/.virtualenvs/django-guardian3/lib/python3.4/site-packages/django/core/management/__init__.py", line 354, in execute
        django.setup()
      File "/home/gkarak/.virtualenvs/django-guardian3/lib/python3.4/site-packages/django/__init__.py", line 21, in setup
        apps.populate(settings.INSTALLED_APPS)
      File "/home/gkarak/.virtualenvs/django-guardian3/lib/python3.4/site-packages/django/apps/registry.py", line 115, in populate
        app_config.ready()
      File "/home/gkarak/.virtualenvs/django-guardian3/lib/python3.4/site-packages/django/contrib/admin/apps.py", line 22, in ready
        self.module.autodiscover()
      File "/home/gkarak/.virtualenvs/django-guardian3/lib/python3.4/site-packages/django/contrib/admin/__init__.py", line 23, in autodiscover
        autodiscover_modules('admin', register_to=site)
[...]

PR #371, opened 2015-12-07
Improving tests in testapp using manage.py

Wtower:
Improve tests using a `manage.py`.
Relevant issue #364


brianmay:
Doesn't seem to work for me without the following change:

```
diff --git a/guardian/testapp/manage.py b/guardian/testapp/manage.py
index bcdef77..2a10a60 100755
--- a/guardian/testapp/manage.py
+++ b/guardian/testapp/manage.py
@@ -6,7 +6,7 @@ if __name__ == "__main__":
     # should be ../../ (`djangoguardian/`)
     BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
-    sys.path.append(BASE_DIR)
+    sys.path.insert(0, BASE_DIR)
 
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "guardian.testapp.testsettings")
```

On my system, `guardian/testapp` is getting inserted at the start of the sys.path, and it looks like this new entry has to appear first.

brianmay:
Was also confused with the title of this pull request: has nothing to do with `example_project`, but `guardian/testapp`.

Wtower:
Ok, I will take a look at that, not today though, rough day. You are right on the title, I just used the issue title but it's completely irrelevant.

Wtower:
I'm back. I was wondering, are you not using a virtualenv? There is for example a [specific question in SO](http://stackoverflow.com/questions/10095037/why-use-sys-path-appendpath-instead-of-sys-path-insert1-path) that recommends use append vs insert, and that this error can be dealt with virtualenv.

Changing the code as you recommended makes no difference to me btw, still running ok so I don't really mind. I was just wondering if this is the correct approach.

Also edited the title.

Wtower:
Also to mention that tests fail for Django 1.7.0, require >=1.7.1 due to 

    TypeError: __init__() got an unexpected keyword argument 'preserve_default'

Which has been dealt in 1.7.1. See [relevant link](http://stackoverflow.com/questions/26432242/django-1-7-migrations).

brianmay:
Made some minor changes and merged this as 6359074a9cbe997cea06a478b19911a15a663795.

Looking at `tests.py` - I see it makes to the settings, which perhaps should get merged into `testsettings.py`.

Not sure there is any need for `tests.py` anymore actually, the `manage.py` looks like a better solution.

Wtower:
Yes, I agree with that. Reduce complexity.

Issue #366, opened 2015-11-27
get_objects_for_user with any_perm=True leads to type error in postgres.


drepo:
This, I believe, is related to `object_pk` being a `VARCHAR(255)`, whereas the `id` fields are usually `INT` types in the `UserObjectPermission` model

A sample of query being generated which is failing (at the `"app_model"."id" IN` clause). The query is relatively longer. I am posting only a part of it.

```sql
SELECT <Fields> FROM "report_report"   
WHERE ("app_model"."id" IN (  
      SELECT U0."object_pk" FROM "guardian_userobjectpermission" U0   
     INNER JOIN "auth_permission" U2 ON ( U0."permission_id" = U2."id" )   
     WHERE (U0."user_id" = 11 AND U2."content_type_id" = 10 AND U2."codename" IN (abc, xyz, efg)))
```

Terminal error:
```bash
psycopg2.ProgrammingError: operator does not exist: integer = character varying
LINE 1: ... FROM "app_model" WHERE ("app_model"."id" IN (SELECT...
                                                             ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

Stack:
Django==1.8.7
django-guardian==1.3.2
python==3.4.3
Linux Mint==17.2 (64-bit)
Postgres==9.3.10

brianmay:
Sorry, not sure I understand what that query is doing or why Postgresql is getting upset.

It would really help if we could see the stack trace, so we can identify what line of code is producing this error.

Basically the above SQ, simplified is:

```
SELECT <Fields> FROM "report_report"  WHERE ("app_model"."id" IN (<list of strings>))
```

Is Postgresql saying it doesn't like comparing app_model.id with a string?

What is app_model and how does it relate to the report_report? What is content_type 10 on your system?

brianmay:
I can see why string is used, I imagine this might be for the case where the id field is not an integer.

In fact, Django code itself does something similar, see the `LogEntry` model in contrib.admin.models.

So I think maybe this might be an upstream Django issue.

brianmay:
Not enough information given to work on. It might be a Django issue, however there is not enough information to report to Django either.

tomka:
[...]

PR #382, opened 2015-12-29
Added method to prefetch the permissions for an iterable of objects 

keattang:
The permissions are then stored in the cache.

This helps reduce the number of database calls required when getting the permissions on a list of objects.


ad-m:
Thanks for you code. It looks very promising and interesting :smile_cat: :+1:

What do you think about tests & docs?

guardian/core.py
```diff
@@ -123,3 +144,78 @@ def get_local_cache_key(self, obj):
...
+            user_perms_qs = model.objects.filter(**user_filters).select_related('permission')
+            group_perms_qs = group_model.objects.filter(**group_filters).select_related('permission')
+            perms = chain(user_perms_qs, group_perms_qs)
+        else:
+            perms = chain(
+                *perms_qs.filter(**group_filters).select_related('permission')
+            )
+
+        for perm in perms:
+            key = (ctype.id, int(perm.object_pk))
```
ad-m: I would like mark that in line 167 key you generate by ``` (ctype.id, pk)```. I'm not sure this is correct to cast ```object_pk``` to int. CharField can have set ```primary_key=True``, so it will be ```object_pk```. In this case it will raises exceptions.

guardian/core.py
```diff
@@ -11,6 +12,26 @@
...
+    if isinstance(objects, QuerySet):
+        model = objects.model
+        pks = list(objects.values_list('pk', flat=True))
+        ctype = ContentType.objects.get_for_model(model)
+    else:
+        for idx, obj in enumerate(objects):
+            if not idx:
+                model = type(obj)
+                ctype = ContentType.objects.get_for_model(model)
+            pks += obj.pk
```
ad-m: Undefined name 'pks'. Add ```pks = [] ``` after line 25.

ad-m:
I want to drop here this snippet to take awareness about ```Prefetch```:
```
from django.db.models import Prefetch
(…)
from guardian.models import GroupObjectPermissionBase, UserObjectPermissionBase
(…)
class CaseQuerySet(QuerySet):
    def with_involved_staff(self):
        qs = (CaseUserObjectPermission.objects.filter(user__is_staff=True).
              select_related('permission', 'user').
              all())
        return self.prefetch_related(Prefetch('caseuserobjectpermission_set', queryset=qs))
(…)
class Case(models.Model):
(…)
    name = models.CharField(max_length=150, verbose_name=_("Subject"))
(…)
class CaseUserObjectPermission(UserObjectPermissionBase):
    content_object = models.ForeignKey(Case)

```

keattang:
I've added fixes for the two issues you pointed out. Do they look ok?

I'll add some tests and docs in the coming days. Where is the best place to add the docs? Directly in the repo?

I'll take a closer look at the prefetch solution but I'm not sure it will work for my use case.

ad-m:
> Where is the best place to add the docs? Directly in the repo?

Yes, in ```/docs/```. It will auto-rebuild for rtfd (see ```repository``` section in http://readthedocs.org/projects/django-guardian/ ).

> (…) I'm not sure it will work for my use case.
I would like mark that I use ```with_involved_staff``` to show grouped users (and perms) who have access to cases in ```CaseListView```, not for permission check. I note only to think about, but I'm not sure. I don't use groups, models perms etc.

brianmay:
At the moment my priority is fixing #366.

In the meantime however you might want to compare with other proposals for improving the performance: https://github.com/django-guardian/django-guardian/labels/Performance

keattang:
No worries. This is ready to go whenever you're ready. I've added tests which are passing and documentation. I'm not sure why the travis build is failing but it doesn't seem to be related to my code.

From what I can see the other issues revolve around getting users for objects while this deals with getting perms for objects.

guardian/testapp/tests/test_core.py
```diff
@@ -181,3 +181,106 @@ def test_get_perms(self):
...
+
+            # Checking for same model but other instance shouldn't spawn any queries
+            checker.has_perm("change_group", new_group)
+            self.assertEqual(len(connection.queries), query_count)
+
+        finally:
+            settings.DEBUG = False
+
+    def test_prefetch_group_perms(self):
+        settings.DEBUG = True
```
ad-m: Maybe I ask stupid question,but...Why do you need this settings update?

keattang: connection.queries only works when django is in debug mode

ad-m:
@keattang, no worries abou Travis build. Everything is fine. See #384.

guardian/testapp/tests/test_core.py
ad-m: Why not ```TransactionTestCase.assertNumQueries```?

keattang: Because you still need to use `connection.queries` to find the initial number of queries to compare against

ad-m: Thank you for your replies.

keattang: No worries :)

brianmay:
How does this compare with #295?

This patch appears to cache things in the ``ObjectPermissionChecker`` object; #295 appears to cache globally per thread.

Not sure how to compare, or if one of these changes becomes redundant now, or what.

keattang:
#295 only seems to only cache the lookup of the object's permissions model
```
_related_object_permission_models_cache.user_obj_perms_models_cache[obj.__name__] = get_obj_perms_model(obj,
                                     UserObjectPermissionBase,
                                     UserObjectPermission)
```
Whereas this patch caches the actual permissions and does so on a list of object rather than a single object so reduce the number of lookups when iterating through the list from `len(obj_list)` to .

Both of these should therefore be compatible and beneficial

keattang:
Any progress on this?

samelliott89:
Any progress on this?

brianmay:
Reviewing this is still on my todo list...

samelliott89:
Any progress on this? Thanks

samelliott89:
Hi @brianmay  we're really looking to use this to make improvements, any chance if can get looked at in the near future? thanks

brianmay:
Rebased, squashed, and merged in e6e45ba81ae0a17172d7a41d4f2c7c274f9215e4.

keattang:
Awesome thanks!

pgeez:
Awesome work @keattang! This is a super helpful feature.

Are you planning on adding support in ```prefetch_perms``` for Direct Foreign Key permissions? There are a few places where references to ```object_pk``` are hardcoded and I was wondering if something similar to this https://github.com/django-guardian/django-guardian/blob/master/guardian/core.py#L89-L95 can be done instead.

PR #383, opened 2016-01-06
Add kwarg to get_users_with_perms to filter for specific permissions

timabbott:
This helps us filter for a list of users who have a particular permission
on an object, not just any permission.

If attach_perms=True is also specified the resultant dictionary value will
still list all the user's permissions, not just the one queried.


mitar:
What is a difference between this pull request and #246?

timabbott:
`only_with_perms` is an iterable, not just a single permission; also this one comes with some tests :)

guardian/shortcuts.py
```diff
@@ -215,6 +223,14 @@ def get_users_with_perms(obj, attach_perms=False, with_superusers=False,
...
             user_filters = {'%s__content_object' % related_name: obj}
         qset = Q(**user_filters)
+        if only_with_perms is not None:
+            permission_qset = Q()
+            for permission in only_with_perms:
+                permission_qset |= Q(**{
+                    '%s__permission' % related_name: Permission.objects.get(
+                        content_type=ctype, codename=permission)
+                    })
+            qset &= permission_qset
```
mitar: You should also restrict permissions on the `groups__%s__permission`. So I had in #246:

```
group_filters.update({
  'groups__%s__permission' % group_rel_name: permission,
})
```

That part seems missing here?

mitar:
I like this pull request, because I also need this feature. I proposed #246, but implementation here also looks OK, if also group permissions will be checked.

I nominate this pull request for merging in after that.

mitar:
@timabbott, can you update the pull request to latest `master`? There are conflicts.

timabbott:
@mitar OK rebased past the merge conflicts and added that missing hunk.  I think we should add a test for the group permissions case you suggested we fix, but am a bit swamped right now; would you be up for writing one?

satyanash:
added test for group permissions in #555

brianmay:
Oops. Meant to review #555, sorry.

brianmay:
If this has been superseded now by #555, we probably should close it.

thclark:
Closing as superceded by the (now-merged) #555

PR #395, opened 2016-02-09
Fix deprecated code for django 1.10 compatibility

ad-m:
Hello,

I prepared some commits to add supports django 1.10 compatibility.  I know django 1.10 aren't releases yet, but I only changes around deprecation warning in django 1.9 (released).

Changes:

1. Drop django.conf.urls.patterns (django 1.10 compat) in reference to https://github.com/django/django/blob/stable/1.9.x/django/conf/urls/__init__.py#L80-L85
1. Add modern settings for templates engine in reference to https://docs.djangoproject.com/en/1.9/ref/templates/upgrading/
1. Drop django.core.management.base.NoArgsCommand in reference to https://github.com/django/django/blob/stable/1.9.x/django/core/management/base.py#L574-L578
1. Drop python 3.3 & django master in Travis (lack of support in django 1.9+)
1. Drop strings view arguments in reference to https://github.com/django/django/blob/stable/1.9.x/django/conf/urls/__init__.py#L103-L108
1. Add template_debug_{setter,getter} in reference to https://docs.djangoproject.com/en/1.9/ref/templates/upgrading/ 
> If it sets TEMPLATE_DEBUG to a value that differs from DEBUG, include that value under the 'debug' key in 'OPTIONS'.

In e1916bb I'm not sure maybe versions depend will be better?

All tests pass in django-master now, even allow failures django master.

Greetings,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,13 @@ matrix:
         env: DJANGO_VERSION=1.9.1 DATABASE_URL=mysql://root:@localhost/django_guardian
       - python: 3.3
         env: DJANGO_VERSION=1.9.1 DATABASE_URL=sqlite://
+      # Drop python 3.3 and django master
+      - python: 3.3
+        env: DJANGO_VERSION=master DATABASE_URL=postgres://postgres@/django_guardian
+      - python: 3.3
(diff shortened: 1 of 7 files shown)


brianmay:
Rebased and applied. Thanks.

PR #400, opened 2016-02-21
Fixed issue for custom user model with custom USERNAME_FIELD

### LINKED ISSUE #399: Cannot create Anonymous user when using a custom user model
If using a custom user model with no `username` attribute, creation of the `AnonymousUser` will fail.

rhblind:
Fixed #399


rhblind:
Whops need to support older versions as well

brianmay:
Thanks for this. Have merged.

Just realized none of this is documented, something that should get fixed too I guess.

rhblind:
No problem.
Thanks for a great library!

Issue #263, opened 2014-09-10
Admin panel filter list lists all objects rather than owned objects


HEdingfield:
When using the filter feature in the admin panel, it shows all objects in the database, rather than just the ones owned by the user. Of course, if the user clicks a filter for something they don't own, the change_list is empty, which is good, but it would be far better if the user only saw options in the dropdown box that related to objects they owned in the filter list.

```
class MyModelAdmin(GuardedModelAdmin):
    user_owned_objects_field = 'creator'
    user_can_access_owned_objects_only = True

    list_display = ('question', 'creator', 'opened')
```

e.g. in the above example, you can see all "questions" and "creators" for all objects in the DB, not just the ones related to the MyModel you own.

mitar:
What does "option" in "shows all options in the database" means to you?

HEdingfield:
@mitar clarified via edit. By "option" I meant the entries that appear in the dropdown box when you're filtering, i.e. objects in the database.

mitar:
Can you maybe make a screenshot? I think I understand but I am not sure. You just dislike that there are filtering options which have no entries for a particular user?

But I think you would have to implement your own filtering widget to limit options based on permissions.

HEdingfield:
Let me try to give an example:

I'm saying if I'm using the Django admin backend looking at, say, a bunch of votes on a survey I created (votes and surveys each being models, with the former having a foreignkey of the latter), when I drop down the UI element to filter my votes by survey, I should only see surveys that I've created, not every single survey ever created by anyone who's used the site.

As it stands, it's not only inconvenient for the user, but also potentially even a security issue. I.e. a random user shouldn't be able to know about all the surveys ever created on the site.

IMO the "filtering widget" you suggest should come default with django-guardian.

mitar:
> when I drop down the UI element to filter my votes by survey

Which UI element is this?

rhblind:
@HEdingfield This would require the `survey` or `votes` object to have a relation to your user model, which most of the time is not the case.
One way to go is to filter out any objects you don't have permissions on using `get_objects_for_user`, but as Django only ship with `add, change and delete` default permissions, it's hard to say in a general way which ones should be required in order to construct the initial queryset for the dropdown widget.

I think this is one of those cases that does not really fit into the library, because it's hard to implement in a general way, but easy to implement in your own solution.

To get you started:

Say you have some permissions named `view_<model_name>` ie. `view_survey`.

```python
[...]

PR #403, opened 2016-02-25
Be precise about permissions

mitar:
For super users one should display only really assigned permissions and not simply all permissions. It makes it hard in the admin to modify per-object permissions for super users otherwise.

### DIFF
--- a/guardian/admin.py
+++ b/guardian/admin.py
@@ -15,7 +15,8 @@ from django.utils.translation import ugettext, ugettext_lazy as _
 from guardian.compat import OrderedDict, get_user_model, get_model_name
 from guardian.forms import UserObjectPermissionsForm
 from guardian.forms import GroupObjectPermissionsForm
-from guardian.shortcuts import get_perms
+from guardian.shortcuts import get_user_perms
+from guardian.shortcuts import get_group_perms
 from guardian.shortcuts import get_users_with_perms
 from guardian.shortcuts import get_groups_with_perms
 from guardian.shortcuts import get_perms_for_model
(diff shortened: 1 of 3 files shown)


brianmay:
Does this have any connection with #263?

mitar:
I do not think so. That seems something else.

The issue here is that permissions are always returned as effective permissions. So there was no standard way to get just really set permissions, not inferred. This is problem in admin, because you get all permissions checked in the admin, if you are superuser. You get all permissions listed as assigned to the object, but when you try to remove one, you cannot.

Similarly the problem is (was) with `get_users_with_perms`. If you used `attach_perms` it would attach effective permissions (so all, or user + group), even if you disabled getting group of super-user permissions.

brianmay:
Ok, so if I understand this correctly, this doesn't break the current API, it adds new functions and leaves the ``get_perms()`` function behaving in the same way.

mitar:
Yes, it is mostly a refactor, and does not change anything from before, **except** that the admin interface will start showing assigned permissions even for superadmins, and not all possible permissions.

mitar:
Similarly, `get_perms_for_model` and `get_groups_with_perms` when used with `attach_perms` will return really the attached permissions and not all permissions for superusers.

brianmay:
Looks like this patch is good, and probably worth applying, however may not be complete.

For the example_project, If I revoke on my my permissions for the super user, after this change is applied I see the correct results at the following page:

http://localhost:8001/admin/posts/post/1/permissions/

If I click the edit button however, it lists that I have all available permissions.

http://localhost:8001/admin/posts/post/1/permissions/user-manage/2/

There were also some problems with the example project which I will fix now.

brianmay:
Rebased in a6a637829c726d5147f658ef0df6b079b6f56af8

mitar:
Cool!

mitar:
Should the example be updated to Django 1.9 from Django 1.6 (in requirements?).

brianmay:
Sounds like a good idea to me.

mitar:
What other issues have you seen with examples?

brianmay:
Thanks for your work on this, and also on the other related tickets. I don't have time to review your changes now, hopefully I will get a chance tomorrow. I don't think I noticed any problems with the examples, except for what I mentioned above.

PR #408, opened 2016-03-09
should fix #271

### LINKED ISSUE #271: Mixed group and user direction relations broken
The docs say "It is also possible to use only user or only group based direct relation, however it is discouraged (it’s not consistent and might be a quick road to hell from the mainteinence point of view, especially)."

but looking at the code here:
https://github.com/lukaszb/django-guardian/blob/dceb6beb5830746b624a7ac99d93949ff263fb98/guardian/shortcuts.py#L402

it seems that if user is generic and group is not, then the `fields` will be overridden. The same is true in the opposite configuration. That if both should be moved after the `if use_groups` block. It also needs to be copied into the `if not any_perm:` section.

scholtalbers:
@cancan101 is this what you expected?


guardian/core.py
```diff
@@ -128,8 +128,6 @@ def get_perms(self, obj):
         ctype = ContentType.objects.get_for_model(obj)
         key = self.get_local_cache_key(obj)
         if key not in self._obj_perms_cache:
-
```
mitar: Unrelated change?

guardian/testapp/tests/test_direct_rel.py
```diff
@@ -214,3 +215,23 @@ def test_get_users_with_perms_plus_groups(self):
...
+    def test_get_users_with_perms_plus_groups_reverse_mixed(self):
+        User.objects.create_user('john', 'john@foobar.com', 'john')
+        jane = User.objects.create_user('jane', 'jane@foobar.com', 'jane')
+        group = Group.objects.create(name='devs')
+        self.joe.groups.add(group)
+        assign_perm('add_reversemixed', self.joe, self.reverse_mixed)
+        assign_perm('change_reversemixed', group, self.reverse_mixed)
+        assign_perm('change_reversemixed', jane, self.reverse_mixed)
+        self.assertEqual(get_users_with_perms(self.reverse_mixed, attach_perms=True),
+                         {
```
mitar: Indentation?

mitar:
Seems cool.

scholtalbers:
@mitar shall I update matching your comments or wait for more input.

mitar:
Hm, I would update, so that we get closer to the state ready for merging.

scholtalbers:
hmm the indentation actually matches the rest of the file, does it not?

mitar:
> hmm the indentation actually matches the rest of the file, does it not?

Oh, you are right. So strange code style.

brianmay:
Rebased, squashed, and merged  in 7049a57bf90f3234632d1d6a1640e05eb4b87b32.

PR #413, opened 2016-03-12
Fixed union between queries

mitar:
This fixes `When merging querysets using 'or', you cannot have extra(select=…) on both sides` exception.

Instead of doing union of querysets, we do union of queries and then execute filter only once.


brianmay:
Incorporated in 45e1d574241b5881b5d5e2b10e2dd051e7835f11.

PR #414, opened 2016-03-17
Lookup anonymous user using custom username field

gnublade:
Fix error in get_anonymous_user with a custom user model without a username field.


brianmay:
Thanks!

stefanw:
Thanks! I'm using `django-guardian` and a custom user model, so this commit is essential – currently 1.4.2 is breaking my tests. Could you do a PyPI release for this?

brianmay:
Done.

Issue #338, opened 2015-08-05
Same permission name for multiple models


fatal10110:
I have 3 models with same permission name, lets say "edit" when im trying to get objects for user, it raise an exception "_get() returned more than one ContentType -- it returned 3!_" 

example:

```
get_objects_for_user(user, 'app_label.edit', klass=Model1.objects.all())
```

Also there is a conflict with polymorphic models ( django-polymorphic ) we are not able to assign permissions to parent model instance.

Thanks.

brianmay:
@lukaszb Can you please comment on this ticket? Not sure I understand the database schema well enough...

lukaszb:
@fatal10110 it's hard to say at which point it fails without full stack trace, can you provide one? Also,  which Django/Python versions you are using?

As to the second issue, regarding django-polymorphic - as a rule of thumb we are not supporting 3rd party packages that does something with your model, sorry. If someone would want to work on it, feel free to create another package that would bind guardian and polymorphic.

brianmay:
I would assume it is failing on this line
https://github.com/django-guardian/django-guardian/blob/devel/guardian/shortcuts.py#L385

If the app_label and codename for the query are the same for more then one object, then it looks like it will have problems.

Full stack traces would be good however so we don't have to make assumptions.

brianmay:
It occurred to me that there really should be a test case for this.

fatal10110:
https://github.com/chrisglass/django_polymorphic/issues/122

brianmay:
I have a test case for this that reproduces the failure. Will try to link it here. Normally this is not an issue, as usual practice is to encode the model name into the permission. e.g. change_group rather then just change.

brianmay:
Error happens on this line in get_objects_for_user:

```python
new_ctype = ContentType.objects.get(app_label=app_label,
                                    permission__codename=codename)
```

martin-xia:
Having the same problem here, don't think ContentType has the attribute permission...

bcopos:
Also have the same problem. Using `django-rest-framework` with `django-polymorphism` and `django-guardian`. I don't think the correct get_content_type function is called when using Django Guardian and DjangoObjectPermissionsFilter together. Last three entries in stack trace below:

```
python2.7/site-packages/guardian/shortcuts.py in get_objects_for_user -- line 471 -- new_ctype = ContentType.objects.get(app_label=app_label, permission__codename=codename)
[...]

PR #421, opened 2016-04-19
Add support for django-polymorphic models.

benkonrath:
This builds upon the work from @bonniejools in this repo:

https://github.com/Anglobit/django-guardian/tree/master

This has been discussed in the django-polymorphic repo: https://github.com/chrisglass/django_polymorphic/issues/122

I'm not sure if this is complete but I'd like to get some input on how I can improve PR with an eye towards getting it merged. Thanks!

### DIFF
--- a/guardian/ctypes.py
+++ b/guardian/ctypes.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 from django.contrib.contenttypes.models import ContentType
 
 
--- a/guardian/ctypes.py
+++ b/guardian/ctypes.py
@@ -1,0 +1,31 @@
+from django.contrib.contenttypes.models import ContentType
+
+
+def get_ctype_from_polymorphic(obj):
+    """
+    Special helper function to return BASE polymorphic ctype id
+    """
+    if hasattr(obj, 'polymorphic_model_marker'):
+        try:
+            superclasses = list(obj.__class__.mro())
+        except TypeError:
+            # obj is an object so mro() need to be called with the obj.
+            superclasses = list(obj.__class__.mro(obj))
+
+        polymorphic_superclasses = list()
+        for sclass in superclasses:
+            if hasattr(sclass, 'polymorphic_model_marker'):
+                polymorphic_superclasses.append(sclass)
+
+        # PolymorphicMPTT adds an additional class between polymorphic and
+        # base class
+        if hasattr(obj, 'can_have_children'):
+            root_polymorphic_class = polymorphic_superclasses[-3]
+        else:
+            root_polymorphic_class = polymorphic_superclasses[-2]
+        ctype = ContentType.objects.get_for_model(root_polymorphic_class)
+
+    else:
+        ctype = ContentType.objects.get_for_model(obj)
+
+    return ctype
--- a/guardian/backends.py
+++ b/guardian/backends.py
@@ -4,6 +4,7 @@ from django.db import models
 
 from guardian.compat import get_user_model
 from guardian.conf import settings
+from guardian.ctypes import get_ctype_from_polymorphic
 from guardian.exceptions import WrongAppError
 from guardian.core import ObjectPermissionChecker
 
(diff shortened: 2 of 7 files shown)


benkonrath:
I just read through #338 saw this comment from @lukaszb:

> As to the second issue, regarding django-polymorphic - as a rule of thumb we are not supporting 3rd party packages that does something with your model, sorry. If someone would want to work on it, feel free to create another package that would bind guardian and polymorphic.

I don't know another way to override `ContentType.objects.get_for_model(obj)` besides what is being done in the PR. Would you accept a PR that allows third party apps to override `ContentType.objects.get_for_model(obj)` somehow? Instead of the `get_ctype_from_polymorphic` that I have now, I could add a generic `get_ctype` method with a default implementation like this:

```
get_ctype(obj):
    return ContentType.objects.get_for_model(obj)
```

This could be overridden by a setting that takes a callable to return the appropriate ctype. What do you think? I'm open to other suggestions as well.

guardian/ctypes.py
```diff
@@ -0,0 +1,31 @@
+from django.contrib.contenttypes.models import ContentType
+
+
+def get_ctype_from_polymorphic(obj):
```
ad-m: Does this require testing?

benkonrath: Yeah. I was hoping to get general feedback on the approach before completing the full implementation with tests.

brianmay:
Never heard of Django Polymorphic before. It looks interesting.

Can somebody please explain to me how this helps Django Guardian? I don't think I understand the ``get_ctype_from_polymorphic`` function or what purpose it serve.

Thanks for the pull request.

benkonrath:
django-polymorhpic allows models that inherit from a base model to be retrieved by using a queru on the base model model. For example:
```
Project.objects.all()
    [ <Project:         id 1, topic "Department Party">,
      <ArtProject:      id 2, topic "Painting with Tim", artist "T. Turner">,
      <ResearchProject: id 3, topic "Swallow Aerodynamics", supervisor "Dr. Winter"> ]
```
Using a permission on the base model which needs to apply to all child models doesn't work because django-guardian uses the content type for the child model instead of the base model.

I picked up this patch from chrisglass/django_polymorphic#122 so I'm exploring a bit as well.

I tried to get the functionality I need without this patch by making wrapper functions to the various guardian shortcuts. My idea was to supply an instance of the base model instead of the child model when using these shortcut functions. Unfortunately this didn't work. 

Here's an example with `assign_perm`:
```
from guardian.shortcuts import assign_perm

def assign_polymorphic_perm(perm, user_or_group, obj=None):
    if obj:
        base_obj = get_base_polymorphic_instance(obj)
        return assign_perm(perm, user_or_group, base_obj)
    return assign_perm(perm, user_or_group)

assign_polymorphic_perm('projects.can_edit', <Group instance>, <ArtProject instance>)
```

This raises a `ValidationError` because of the way that content types are retrieved:
https://github.com/django-guardian/django-guardian/blob/f46f6702bfae48b46efff1bf1a06f19550699e21/guardian/models.py#L41
```
ValidationError: [u"Cannot persist permission not designed for this class (permission's type is <ContentType: Project> and object's type is <ContentType: ArtProject>)"]
```

I think the best way forward is the suggestion from my previous comment: retrieve the content type in a single place within guardian instead of using `ContentType.objects.get_for_model(obj)` directly everywhere. For example:
```
get_content_type_for_model(obj):
    return ContentType.objects.get_for_model(obj)
```
Third party applications or sites would need a way to override this default function with their custom version. The benefit to django-guardian is that it would allow a third party app to be created that integrates django-polymorphic (and may other libraries) with django-guardian without polluting django-guardian itself. This also allows people to consolidate around a shared solution for this problem instead of the current situation where there are a bunch of maintained forks that people are using and then forking again.

@brianmay Let me know if you like this idea and I'll make a PR for your to provide feedback on. If you're not interested, that's ok too. I'm ok with just using my patched version and closing this PR in this case. Thanks for your work on guardian and for your feedback.

brianmay:
So if I understand you correctly, you want to be able to assign permissions to ``Project`` and have them affect ``ArtProject`` and ``ResearchProject`` too?

The concept sounds good to me.

My understanding of the reading the code is that you will get exactly the same behaviour if the model does not get derived from ``PolymorphicModel``. So it shouldn't break existing code.

It is possible I guess that there is existing code already using ``PolymorphicModel`` that breaks with this new behaviour - how likely is this? I suspect it might be unlikely - however really have no idea.

I guess my main (weak??) concern is that this is somewhat different semantics and people may not expect this significant change just by changing their models to use ``PolymorphicModel``. I guess the way to counter this would be to make sure this is documented in the documentation and easy to find.

benkonrath:
> So if I understand you correctly, you want to be able to assign permissions to Project and have them affect ArtProject and ResearchProject too?

Yeah, that's correct.

> It is possible I guess that there is existing code already using PolymorphicModel that breaks with this new behaviour - how likely is this? I suspect it might be unlikely - however really have no idea.

This is hard to know and leads into the concern you raised. This solution is definitely making a decision that affects the logic of the permissions when using guardian with django-polymorphic. While this solution works for my current project, I could see it being a problem in systems with other requirements. Documentation won't help solve this because there would be way to opt-out of using the base class for the permissions of the child classes (when using django-polymorhpic).

I'm going to implement the solution that I mentioned in a previous comment so that we can see if it's a better way forward for solving this problem. I hope to be able to get to it in the next week or so. Let's keep this PR open until we figure out which solution is best. Thanks for your feedback.

benkonrath:
Closed in favour of #431.

PR #434, opened 2016-06-22
Add example of CRUD CBV application

ad-m:
Hello,

There is a basic example of use django-guardian in FBV-based application.  I suggest add example of effective usage of django-guardian in CBV-based application.

I think that for some people a good example of the use of the library is much easier to understand than the API documentation library. I'm open to suggestions, because I wonder if you can write something smarter. I hope this example will promote the best ways to use the library.

This pull requests depends on #433 .

Greetings,


ad-m:
I added ```PermissionListMixin```. I can add docs to them if useful.

brianmay:
Rebased and merged. ce241f3d87d098bfef5a8774924381ccc3e350bb and 9c00a6e3146ce4c67ca630c18a8c5894c1ba2940.

Thanks for the contribution.

PR #437, opened 2016-07-12
Sorted imports with isort

atombrella:
Appended on_delete=models.CASCADE as this will become mandatory in Django 2

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,4 @@
+[settings]
+line_length=120
+force_single_line=False
+force_alphabetical_sort=True
--- a/guardian/admin.py
+++ b/guardian/admin.py
@@ -1,26 +1,20 @@
 from __future__ import unicode_literals
-
-import django
 from django import forms
 from django.conf import settings
-from guardian.compat import url
-from django.contrib import admin
-from django.contrib import messages
+from django.contrib import admin, messages
 from django.contrib.admin.widgets import FilteredSelectMultiple
 from django.core.urlresolvers import reverse
-from django.shortcuts import render_to_response, get_object_or_404, redirect
+from django.shortcuts import get_object_or_404, redirect, render_to_response
 from django.template import RequestContext
-from django.utils.translation import ugettext, ugettext_lazy as _
-
-from guardian.compat import OrderedDict, get_user_model, get_model_name
-from guardian.forms import UserObjectPermissionsForm
-from guardian.forms import GroupObjectPermissionsForm
-from guardian.shortcuts import get_user_perms
-from guardian.shortcuts import get_group_perms
-from guardian.shortcuts import get_users_with_perms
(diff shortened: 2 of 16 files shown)


brianmay:
Sorry, can't apply this when it breaks *all* the tests. Some sort of Syntax Error (that I can't actually see right now) in the migration:

```
E         File "/home/travis/build/django-guardian/django-guardian/guardian/migrations/0001_initial.py", line 24
E           ('content_type', models.ForeignKey(to='contenttypes.ContentType'), on_delete=models.CASCADE),
E                                                                                       ^
E       SyntaxError: invalid syntax
```

Also think it might be better to create new migrations instead of updating the existing migrations; existing installs won't see these changes otherwise.

brianmay:
Also looks like you have combined two very different changes into one commit. In future would be appreciated if you separate them into two distinct commits.

One that changes the sort order of the imports (why?) and the other that adds the on_delete parameter.

atombrella:
on_delete=models.CASCADE

is already default behavior and there isn't need to create a new migration for it (the warning highlights this, but it's mandatory to explicit state it in future releases of django.

Changing the order of imports appear to be good style; at least a lot of other big projects, e.g., django use this practice.

I can create a new PR for just the `on_delete` adjustment, and disregard the import sorting for now?

ad-m:
I agree with @atombrella . 

It's a good idea to update migrations than make a new migrations. No database update is required databases for current use, but migrations update is required to Django 2 compatibility. @atombrella , do you can update migrations to remove syntax errors?

@brianmay, see django 1.9 releases note draft ( https://docs.djangoproject.com/en/dev/releases/1.9/#foreignkey-and-onetoonefield-on-delete-argument ):
> In order to increase awareness about cascading model deletion, the on_delete argument of ForeignKey and OneToOneField will be required in Django 2.0.
> 
> Update models and existing migrations to explicitly set the argument. Since the default is models.CASCADE, add on_delete=models.CASCADE to all ForeignKey and OneToOneFields that don’t use a different option. You can also pass it as the second positional argument if you don’t care about compatibility with older versions of Django.

It's a good idea to sort imports too. The code is a more predictable what's important for merge (add import to two side, but in different line make double import after merge, so it's import to add in predictable line every time), easier to read and verify import existing etc.). 

We have to consider use in ```isort``` settings "set force_single_line to True (-sl on the command line) and every import will appear on its own line". It is current practice in this project and I think that there is no reason to change this now. I might be better to specify project level configuration simply by placing a ```.isort.cfg``` file at the root of our project with that settings.

@brianmay , it's a good idea to update your editor to use ```isort``` if you like. There is a lot of plugins.

brianmay:
Ok, I am going to merge this. In future however it would be better if you could keep distinct changes as distinct commits. Thanks.

brianmay:
Merged as f37746a11a451ca72890ca1ca826eaff940ae8a8.

PR #438, opened 2016-07-13
Make UserManage form overridable

julianandrews:
This allows simpler user select forms (for instance a `ModelChoiceField`).

It's a fairly simple change, but I think being able to customize the `UserManage` form without monkey-patching, or copy-pasting the whole method is probably a good idea.

### DIFF
--- a/guardian/admin.py
+++ b/guardian/admin.py
@@ -178,8 +178,8 @@ class GuardedModelAdminMixin(object):
         )
 
         if request.method == 'POST' and 'submit_manage_user' in request.POST:
-            user_form = self.get_obj_perms_user_select_form()(request.POST)
-            group_form = self.get_obj_perms_group_select_form()(request.POST)
+            user_form = self.get_obj_perms_user_select_form(request)(request.POST)
+            group_form = self.get_obj_perms_group_select_form(request)(request.POST)
             info = (
                 self.admin_site.name,
                 self.model._meta.app_label,
@@ -193,8 +193,8 @@ class GuardedModelAdminMixin(object):
                 )
                 return redirect(url)
         elif request.method == 'POST' and 'submit_manage_group' in request.POST:
-            user_form = self.get_obj_perms_user_select_form()()
-            group_form = self.get_obj_perms_group_select_form()(request.POST)
+            user_form = self.get_obj_perms_user_select_form(request)()
+            group_form = self.get_obj_perms_group_select_form(request)(request.POST)
             info = (
                 self.admin_site.name,
                 self.model._meta.app_label,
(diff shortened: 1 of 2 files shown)


julianandrews:
I made the same change for the GroupManage form for much the same reasons.

tdruez:
+1 for this, very clean and useful change!

@julianandrews would you mind to add the request object in the `get_obj_perms_manage_user_form ` and`get_obj_perms_user_select_form()` method parameters?
The same way it's provided in django `ModelAdmin.get_form()`, this would allow, when needed, all kind of form customization using values from the request object.

```
def get_obj_perms_user_select_form(self, request):
    [...]

def get_obj_perms_group_select_form(self, request):	
    [...]
```

Thanks again for your contribution :)

julianandrews:
@tdruez That seems like a reasonable change to me. For consistency, I made the same change in the `get_obj_perms_manage_user_form` and `get_obj_perms_manage_group_form`.

tdruez:
@julianandrews thanks! 
Your changes looks great, this is exactly the kind of flexibility I needed.
Looks like there's a couple tests to fix though ;)

guardian/admin.py
```diff
@@ -178,8 +178,8 @@ def obj_perms_manage_view(self, request, object_pk):
         )
 
         if request.method == 'POST' and 'submit_manage_user' in request.POST:
-            user_form = UserManage(request.POST)
-            group_form = GroupManage()
+            user_form = self.get_obj_perms_user_select_form(request)(request.POST)
+            group_form = self.get_obj_perms_group_select_form(request)(request.POST)
```
brianmay: Here you modified the ``group_form`` constructor call to pass the ``request.POST`` parameter - was this intended?

guardian/admin.py
```diff
@@ -193,8 +193,8 @@ def obj_perms_manage_view(self, request, object_pk):
                 )
                 return redirect(url)
         elif request.method == 'POST' and 'submit_manage_group' in request.POST:
-            user_form = UserManage()
-            group_form = GroupManage(request.POST)
+            user_form = self.get_obj_perms_user_select_form(request)()
+            group_form = self.get_obj_perms_group_select_form(request)(request.POST)
```
brianmay: Here you didn't modify the ``user_form`` constructor call to pass the ``request.POST`` parameter. Guessing that this should be consistent with the above change?

julianandrews:
@brianmay I've added the missing `request.POST` parameter. Yeah, I think there's no reason not to instantiate the form from the `POST` in these spots. It shouldn't have any effect on the default form, but is probably the appropriate behavior in the cases where it matters.

brianmay:
Merged in b82ead3ff61a5931e25a7a4d4e1db2e5760bf4d8.

PR #439, opened 2016-07-15
Upgraded Grappelli templates breadcrumbs block to new Django 1.9 and …

tdruez:
…Grappelli 2.8 standards, including proper URLs and support for preserved_filters. Removed the duplicated field.errors in the field.html template file.


brianmay:
Have no idea if this is correct, however it looks good, so I merged it.

tdruez:
@brianmay thanks for the merge!

See https://github.com/sehmaschine/django-grappelli/blob/master/grappelli/templates/admin/change_form.html#L143 
and https://github.com/django/django/blob/master/django/contrib/admin/templates/admin/change_form.html#L16

The URLs in those template are proper now URL reversing instead of old hardcoded "../../../../".
My changes are about being consistent with those new standards ;)

brianmay:
Yes, noted the urls. Very good. Thanks for the pull request.

Issue #442, opened 2016-08-05
Django 1.10 compatibility for GuardedModelAdminMixin


tdruez:
From https://docs.djangoproject.com/en/1.10/releases/1.10/#features-removed-in-1-10

> The dictionary and context_instance parameters for the following functions are removed:
django.shortcuts.render()
django.shortcuts.render_to_response()
django.template.loader.render_to_string()

See also related https://github.com/django-guardian/django-guardian/issues/370

`render_to_response` in
- 1.10: https://github.com/django/django/blob/stable/1.10.x/django/shortcuts.py#L16
- 1.9: https://github.com/django/django/blob/stable/1.9.x/django/shortcuts.py#L27

The following lines needs to be fixed in django-guardian:
- https://github.com/django-guardian/django-guardian/blob/devel/guardian/admin.py#L217
- https://github.com/django-guardian/django-guardian/blob/devel/guardian/admin.py#L268
- https://github.com/django-guardian/django-guardian/blob/devel/guardian/admin.py#L341

The `RequestContext(request)` vars are provided as the `content_type` argument of the new `render_to_response` function. This breaks the rendering of the view.

brianmay:
Assuming this is closed now that I merged #443. If not, please let me know.

PR #443, opened 2016-08-05
Fixed GuardedModelAdminMixin views render for Django 1.10

tdruez:
#442

### DIFF
--- a/guardian/admin.py
+++ b/guardian/admin.py
@@ -4,7 +4,7 @@ from django.conf import settings
 from django.contrib import admin, messages
 from django.contrib.admin.widgets import FilteredSelectMultiple
 from django.core.urlresolvers import reverse
-from django.shortcuts import get_object_or_404, redirect, render_to_response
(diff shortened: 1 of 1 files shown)


brianmay:
Thanks for this. Probably should add Django 1.10 to the Travis tests at some stage too...

tdruez:
Thanks for merging this, would you mind to issue a new release as well?
The current setup.py (non explicit version) makes it hard to use a non-tagged commit.
Thanks ++

pombredanne:
I am facing the same issue. My head was spinning while trying to figure out how a simple setup.py versions was guessed .... Could there be a simpler, more explicit way that would be friendly to forks and temp dev releases? like using `version="1.4.5"` in `setup.py` .... eventually using `bumpversion` to propagate that in multiple places explicitky?

AlexGrs:
Hi. 

Getting a release for this would be great for those having Django 1.10. 

Thank you.

brianmay:
Version 1.4.5 has been released.

PR #447, opened 2016-08-12
Removed some confusion from the docs. Fixes #446

### LINKED ISSUE #446: Make the fact that "django-guardian" is actively maintained more clear
Hello,

I was looking around for a row-level (object level) permission framework for Django and found django guardian through google searches and this presentations from PyCon 2016: https://www.youtube.com/watch?v=C0wuTkS93B0

However at first I was under the impression that it isn't maintained for any longer due to:

- the [documentation](http://django-guardian.readthedocs.io/en/latest/) saying "Copyright 2010-2013"
- the [overview page](http://django-guardian.readthedocs.io/en/latest/overview.html) talking about Django 1.2

Fortunately I saw on DjangoPackages.org that a lot of people use it and took a second look and realized that it is in fact actively maintained and a version was released just earlier this month. A couple of small documentation changes could such misunderstandings in the future.

PS. In the end I ended up going with the "rules" package since my use case is better suited to that (I don't need to create arbitrary links "this-user-can-access-that-object" but rather have clear rules like "every user can access the objects they own").

lukaszb:


ad-m:
LGTM

PR #460, opened 2016-09-15
Fix django 1.10 deprecation warnings

rpkilby:


guardian/testapp/migrations/0001_initial.py
```diff
@@ -1,106 +1,155 @@
...
+# Generated by Django 1.10.1 on 2016-09-16 11:15
 from __future__ import unicode_literals
 
 import datetime
-
 from django.conf import settings
-import django.core.validators
-from django.db import models, migrations
+import django.contrib.auth.models
+import django.contrib.auth.validators
```
ad-m: ```django.contrib.auth.validators``` was introduced in django 1.10.1 (see https://github.com/django/django/commit/526575c64150e10dd8666d1ed3f86eedd00df2ed ). It's unacceptable to drop support for django<1.10.1. Support for django 1.9 from django foundation has not ended even. Mosts builds fails.

rpkilby: > It's unacceptable to drop support for django<1.10.1. Support for django 1.9 from django foundation has not ended even. 

Agreed - definitely not trying to drop supported versions of Django. The migrations are being difficult to update. On the one hand, you have deprecations which prevent you from using `python -W error` (errors are raised before the test suite even runs). On the other hand, you have compatibility issues (such as the validators from `contrib.auth`).

ad-m:
@rpkilby , thank you for your contribution.

PR #463, opened 2016-10-12
Reduce memory usage of clean_orphan_obj_perms

PierreF:
We use generic FK for your permission and don't yet (shame on us :)) delete permission object when object is deleted.

Therefor we use from time to time clean_orphan_obj_perms... but it iterate over *ALL* permission objects, and we have quiet a lots of them (about 100k)... this cause clean_orphan_obj_perms to cache all objects and consume way to much memory (multiple Gb).

This is a known issue of Django with a known solution: use [iterator](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#iterator).

This PR use iterator() for clean_orphan_obj_perms when iterating over all permission objects.

### DIFF
--- a/guardian/utils.py
+++ b/guardian/utils.py
@@ -128,8 +128,8 @@ def clean_orphan_obj_perms():
 
     deleted = 0
     # TODO: optimise
-    for perm in chain(UserObjectPermission.objects.all(),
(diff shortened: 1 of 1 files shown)


lukaszb:
Thanks, that definitely makes sense!

Issue #292, opened 2015-01-21
Instance Level groups


wldcordeiro:
I'm not sure if this is possible from the documentation but is it possible to add groups that only act on specific instances? For example, you have a Company model with instance level permissions and then you have groups called CompanyUser, CompanyAdmin and give those users in the groups only access to the company instance they're associated with? So if you we had Company A and Company B, the CompanyAdmins of A can't modify B and vice verse.

v-erena:
That is possible. In django console this would translate to: 
```python
>>> from django.contrib.auth.models import User, Group
>>> from guardian.shortcuts import assign_perm
# fictional companies
>>> companyA = Company.objects.create(name="Company A")
>>> companyB = Company.objects.create(name="Company B")
# create groups
>>> companyUserGroupA = Group.objects.create(name="Company User Group A")
>>> companyUserGroupB = Group.objects.create(name="Company User Group B")
# assign object specific permissions to groups
>>> assign_perm('change_company', companyUserGroupA, companyA)
>>> assign_perm('change_company', companyUserGroupB, companyB)
# create user and add it to one group for testing
>>> userA = User.objects.create(username="User A")
>>> userA.groups.add(companyUserGroupA)
>>> userA.has_perm('change_company', companyA)
True
>>> userA.has_perm('change_company', companyB)
False
>>> userB = User.objects.create(username="User B")
>>> userB.has_perm('change_company', companyA)
False
>>> userA.has_perm('change_company', companyB)
True
```

lukaszb:
@v-erena thanks for that. Do you guys want to include similar example to documentation?

wldcordeiro:
@lukaszb That would definitely be nice. I think it's likely a common use case.

PR #477, opened 2016-12-15
Fix group permission example

caymard:
In the documentation I found a little error in one of the [example](https://django-guardian.readthedocs.io/en/stable/userguide/assign.html#for-group). It was introduced with #292.


ad-m:
@caymard , thanks!

Issue #478, opened 2016-12-17
Sample Project


hvilela:
I just download the project and tried to run the sample project following these steps:
https://django-guardian.readthedocs.io/en/stable/userguide/example_project.html

Got these erros:

```
MacBook-Pro-de-Henrique-2:example_project hvilela$ ./manage.py runserver
Performing system checks...

System check identified some issues:

WARNINGS:
?: (1_8.W001) The standalone TEMPLATE_* settings were deprecated in Django 1.8 and the TEMPLATES dictionary takes precedence. You must put the values of the following settings into your default TEMPLATES dict: TEMPLATE_DIRS, TEMPLATE_CONTEXT_PROCESSORS, TEMPLATE_DEBUG, TEMPLATE_LOADERS.

System check identified 1 issue (0 silenced).
December 17, 2016 - 17:48:14
Django version 1.9.4, using settings 'settings'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
Internal Server Error: /
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/django/core/handlers/base.py", line 174, in get_response
    response = self.process_exception_by_middleware(e, request)
  File "/Library/Python/2.7/site-packages/django/core/handlers/base.py", line 172, in get_response
    response = response.render()
  File "/Library/Python/2.7/site-packages/django/template/response.py", line 160, in render
    self.content = self.rendered_content
  File "/Library/Python/2.7/site-packages/django/template/response.py", line 137, in rendered_content
    content = template.render(context, self._request)
  File "/Library/Python/2.7/site-packages/django/template/backends/django.py", line 95, in render
    return self.template.render(context)
  File "/Library/Python/2.7/site-packages/django/template/base.py", line 204, in render
    with context.bind_template(self):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/Library/Python/2.7/site-packages/django/template/context.py", line 260, in bind_template
    updates.update(processor(self.request))
  File "/Users/hvilela/Downloads/django-guardian-devel/example_project/core/context_processors.py", line 5, in version
    return {'version': guardian.get_version()}
AttributeError: 'module' object has no attribute 'get_version'
[17/Dec/2016 17:48:23] "GET / HTTP/1.1" 500 83849
```

ad-m:
[...]

Issue #425, opened 2016-05-06
Incompatible setup_requires setting in setup.py


jcmcken:
As it stands, the ``setup_requires`` setting in ``setup`` is largely incompatible with how ``pip`` handles dependency resolution when performing installations. See pypa/pip#1820, pypa/pip#769, pypa/pip#410. While there are some suggested workarounds and hacks to get it to work, they may or may not work depending on exactly how you're doing the installation.

From what I can tell, these setup dependencies are just used to write out the version file, and so it would seem to be a pretty simple alternative to just do that explicitly without any setup dependencies. What do you think?

brianmay:
Using setup_requires is the recommended way of using setuptools-scm. Straight from their documentation: https://pypi.python.org/pypi/setuptools_scm/1.0.0

setuptools-scm does more then just write the version file, it also determines the correct version to use based on information from git and other sources such as PKG_INFO. It really is required.

Why not just make sure setuptools-scm is installed before calling ``setup.py``?

brianmay:
No response; closing. Can reopen if required.

PR #479, opened 2016-12-18
Replace setuptools_scm into bumpversion

ad-m:


ad-m:
As mentioned in #478 I replace setuptools_scm to bumpversion. I am open to discuss that. It fix #445 . There is also some other issues related to setuptools_scm integration (see #425).

PR #488, opened 2017-02-19
fixes #487 - fix for out of date get_objects_for_user docstring

### LINKED ISSUE #487: Incorrect documentation for guardian.shortcuts.get_objects_for_user
The documentation at https://django-guardian.readthedocs.io/en/stable/api/guardian.shortcuts.html#get-objects-for-user mentions:

`If perms is an empty list, then it returns objects for which a given user has any object permission.`

Unfortunately, providing an empty lists results in a WrongAppError:
`raise WrongAppError("Cannot determine content type")`

It would be very useful to be able to get all objects managed by guardian where a user has permissions. Is this just a bug or has this feature been removed?

Thanks!

mikeengland:


ad-m:
@mikeengland , thanks!

PR #491, opened 2017-03-07
Added a optional checker variable to get_obj_perms

hetsch:
While reading through the documentation I stumbled over `ObjectPermissionChecker.prefetch_perms` which is pretty handy if working in `DetailViews`or  `ListViews` and larger querysets. These modifications work pretty well for me if I instantiate the ObjectPermissionChecker in a view (`ObjectPermissionChecker.prefetch_perms(queryset)`) and then passing the checker instance via the view's context to the template. This saves some queries compared to the current code that doesn't facilitate `prefetch_perms` in https://github.com/django-guardian/django-guardian/blob/devel/guardian/templatetags/guardian_tags.py#L45.

This is just a rough idea with basic testing because I'm not quite sure if I've misunderstood something or if I've overlooked an existing feature or workaround that achieves the same. So feel free to delete this pull request if it makes no sense to you. By the way, thank you for this library, really enjoying it!

### DIFF
--- a/guardian/templatetags/guardian_tags.py
+++ b/guardian/templatetags/guardian_tags.py
@@ -19,10 +19,11 @@ register = template.Library()
 
 class ObjectPermissionsNode(template.Node):
 
-    def __init__(self, for_whom, obj, context_var):
+    def __init__(self, for_whom, obj, context_var, checker=None):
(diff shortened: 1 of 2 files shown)


ad-m:
Thanks!

PR #496, opened 2017-04-01
Added on_delete to all ForeignKeys

michael-k:
`on_delete` is a [required argument beginning with Django 2.0](https://docs.djangoproject.com/en/1.10/ref/models/fields/#foreignkey) (the current master).  Among others this causes the tests against Django's master to fail.

This PR is not enough to fix the tests against Django's master, but it's a start.


ad-m:
@michael-k , thank you.

PR #500, opened 2017-04-07
Exclude Django master + python < 3.5 from test matrix

michael-k:
Django 2.0 requires Python 3.5+
Ref https://docs.djangoproject.com/en/dev/releases/2.0/


ad-m:
@michael-k , thanks!

PR #504, opened 2017-05-10
Custom User Group model support

yonkshi:
We added support for custom user group models. (In our case, we use a custom group model for permission inheritance) 

a new config setting: `GUARDIAN_CUSTOM_USER_GROUP` will load the custom user group module. Similar to a custom User Model, a custom User Group must have all the relationships as Django's built in group

### DIFF
--- a/benchmarks/run_benchmarks.py
+++ b/benchmarks/run_benchmarks.py
@@ -27,11 +27,11 @@ from benchmarks import settings
 from guardian.shortcuts import assign_perm
 from django.core.exceptions import ImproperlyConfigured
 from utils import show_settings
-from django.contrib.auth.models import User, Group
+from django.contrib.auth.models import User
 from django.utils.termcolors import colorize
 from benchmarks.models import TestModel
 from benchmarks.models import TestDirectModel
-from guardian.models import UserObjectPermission
+from guardian.models import UserObjectPermission, Group
 from django.contrib.contenttypes.models import ContentType
 
 USERS_COUNT = 50
--- a/guardian/compat.py
+++ b/guardian/compat.py
@@ -34,7 +34,6 @@ __all__ = [
 # to get_user_model deferred to execution time
 
 user_model_label = getattr(settings, 'AUTH_USER_MODEL', 'auth.User')
-
 try:
     from django.contrib.auth import get_user_model
 except ImportError:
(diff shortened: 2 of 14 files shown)


Merenon:
Exactly what we're searching for. What is the current status / what are the remaining issues?

lukaszb:
@yonkshi thanks, it looks fine. Can you extend documentation? Do you think about something else here (asking as there is still `WIP` in the title)?

yonkshi:
Sorry for the lack of activities lately. The PR should be at a working state, but it probably needs a bit more testing. After submitting the PR, we ended up deferring this issue internally and this custom user group PR was not used in our production. 

@Merenon Feel free to use it, there are no outstanding issues. If you end up adopting this PR, please keep us posted

@lukaszb  Yes, I can definitely extend documentation. Do you want me to add the documentation in the manual or simply more comments inside the code?
As for the WIP, there are no more things I needed to add, I was simply waiting on when I (or somebody else) would come around and test it. However, if you think this is sufficient, I could remove the WIP and merge in once the documentation is ready.

lukaszb:
@yonkshi docs at manual would be awesome - this and a test seem to be the only pieces before we can merge this.

gabrielbiasi:
Any news about this one?

gabrielbiasi:
It seems that although it is possible to implement a custom group (creating a process similar to the User model with the swapper), we have a serious problem in the backend of the auth app.

```
# django.contrib.auth.backends, line 36

def _get_group_permissions(self, user_obj):
    user_groups_field = get_user_model()._meta.get_field('groups')
    user_groups_query = 'group__%s' % user_groups_field.related_query_name()
    return Permission.objects.filter(**{user_groups_query: user_obj})
```

We have "**group**__%s", forcing the foreign field to have the ``related_query_name='group'`` setting to work correctly with the permissions.

While this string is not configurable, we can not have customizable groups (or we have to override the ``_get_group_permissions`` method to allow customization)

Ticket opened in the tracker related to this problem: https://code.djangoproject.com/ticket/29748

ameyapathare:
I have been looking for a way to add a tenant and description field to the Groups model, but still have it work with django-guardian, and this looks promising.  If I make sure that my CustomUser groups is defined with a related_query_name = group, should everything work?

```
class CustomGroup(Group):
    description = models.CharField(max_length=500)
    tenant = models.ForeignKey('accounts.Tenant', on_delete=models.CASCADE)
    users = models.ManyToManyField(CustomUser, related_query_name='group', related_name='groups')

class CustomUser(AbstractUser, GuardianUserMixin):
    tenant = models.ForeignKey('accounts.Tenant', on_delete=models.CASCADE)
```

BonaFideIT:
We have found a better alternative for using a Custom Group model, look at our PR: https://github.com/django-guardian/django-guardian/pull/825/.

We think this PR can be closed since it does not provide a good solution for the problem.

BonaFideIT:
Since we are presenting a working solution for a custom group model and this PR is old and still WIP and has not been updated for some time, we are closing the PR with reference to #825.

PR #521, opened 2017-09-07
Replaced syncdb with makemigrations + migrate

michael-k:
`makemigrations` + `migrate` replaced `syncdb` in Django 1.7, which is no longer supported.
https://docs.djangoproject.com/en/1.7/topics/migrations/


ad-m:
Thanks!

PR #525, opened 2017-09-26
Optymize method save_obj_perms.

nerosketch:
Now to be changed only rows that has been different from value in database. For example you open the form and just save it, logically that nothing to save, but old code removes not selected rows and adds selected, even if values not changed. My commit fix it.


ad-m:
Looks good. Thanks!

PR #526, opened 2017-10-07
Change accept global perms default

italomaia:
Changes default value for accept_global_perms in shortcuts get_object_for_user and get_objects_for_group to False.

### DIFF
--- a/guardian/shortcuts.py
+++ b/guardian/shortcuts.py
@@ -348,7 +348,7 @@ def get_groups_with_perms(obj, attach_perms=False):
 
 
 def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=False,
-                         with_superuser=True, accept_global_perms=True):
+                         with_superuser=True, accept_global_perms=False):
     """
     Returns queryset of objects for which a given ``user`` has *all*
     permissions present at ``perms``.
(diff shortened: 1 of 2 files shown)


guardian/shortcuts.py
```diff
@@ -348,7 +348,7 @@ def get_groups_with_perms(obj, attach_perms=False):
 
 
 def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=False,
-                         with_superuser=True, accept_global_perms=True):
+                         with_superuser=True, accept_global_perms=False):
```
ad-m: I suggest add some dummy object and add warning if not defined explicte. This library is important in terms of security and we can not surprise its users with changes in access policies. I think it is better to introduce a warning now and force an adjustment than to be surprised by the need to adjust.
```
 def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=False,                         with_superuser=True, accept_global_perms=DEFAULT_VALUE):
     if accept_global_perms is DEFAULT_VALUE:
          warnings.warn("The behavior of 'accept_global_perms' is reversed in django-guardian v 2.x. Define the desired behavior directly to maintain compatibility and security in the future.") 
          accept_global_perms = True  # TODO: changed in django-guardian v 2.x
```

RainshawGao: if users used the default value before, and make their code work. then after grade, the code may break. In this case, we should raise warnings.
but if users used a specific value before, after grade, the code could also work. we should not give warnings.
we can make these functions receive a dict, and get values from it.

code maybe like this:
```
def get_objects_for_user(user, perms, **kwargs):
    # get values from kwargs
    ....
```

thclark:
These change suggestions haven't been resolved in 8 years, and I don't feel comfortable making such a significant change to the API in a mature library so I'm closing this one unless someone opens up an issue with a really clear use case for why we should make such a change.

PR #527, opened 2017-10-09
Add Django 2.0 to test matrix

michael-k:


ad-m:
While there is no official 2.0 release I think I should wait. As soon as django released version 2.0 I can merge that.

ad-m:
@michael-k , do you can fix Django version specification?

ad-m:
@michael-k , do you can fix django version classifiers in setup.py?

ad-m:
Fix in #537

PR #529, opened 2017-10-20
Filter attached perms by ctype in get_groups_with_perms

dblenkus:
``get_groups_with_perms`` called with ``attach_perms=True`` doesn't filter permissions by content type, so permissions for all objects with the same ``pk`` are returned instead of just permissions for given object.


ad-m:
Do you can add test for that?

dblenkus:
I've added a test that fails on master and passes in this pull request.

PR #537, opened 2017-12-03
Add django 2.0 to tests & update docs 

ad-m:


ad-m:
Since Django 2.0 is not available in Python 2.7, and django-environ automatically installs the latest version of Django, we must install all dependencies in one command to ensure correct version resolution by pip.

Issue #380, opened 2015-12-22
Global permissions vs object permissions


brianmay:
There seems to be a number of bugs due to some routines using global permissions, and others only using object permissions. Will close the related bugs and reference them here.

Pesticles:
Yes to this!

permission_required_or_403 has the accept_global_perms parameter, but there's no way in a view to do the same thing with get_obj_perms

I propose a get_obj_perms_with_global tag which includes global perms.

I don't know if you've already started down this road, but I'll start working up a patch to do it; if you've already made a start feel free to stop me :)

P.S. I think there's an argument to be made that the default behaviour should be to include global permissions, it seems to me that when you ask "can user X do Y to object Z" that implies checking all of user/object, group/object, user/global, group/global permissions. But that might just be because of the the way I use it.

Pesticles:
Having said all that, it is pretty straight forward to check global permissions separately...Hrmmm.

aidanlister:
Pesticles: there's enough issues on guardian as testament to how confusing this is. We get bit by it every six months ... if request.user.has_access('foo', obj) should be true if they have global permissions but not object level permissions. Writing the check twice is yucky.

What needs to happen to get this changed?

brianmay:
I think somebody needs to write a pull request... Ideally in such a way that it doesn't break existing code (if this is sane and feasible or not I don't know).

rvk86:
Old issue, but I don't see a pull request yet. Any news on this one? I'm running into the same behavior, which I think is not very predictable. Thanks a bunch!

doganmeh:
I added some comments to #49 since that had more insight on the issue of whether and how to fallback to globals when object level fails. I am not sure if closed issues raise notification, so I am adding this here.

doganmeh:
Depending on the design decisions being made I have intent to work on some of these issues.

airstandley:
I would be happy to help with these issues.

doganmeh:
#327 seemed fixed. 

```
brandon=Person.objects.get(first_name='BRANDON')
brandon.is_superuser
False

school=School.objects.get(pk=1)

get_users_with_perms(school, attach_perms=True, with_superusers=False, with_group_users=False)
{}

assign_perm('add_school', brandon, school)
<UserObjectPermission: UserObjectPermission object (1)>

get_users_with_perms(school, attach_perms=True, with_superusers=False, with_group_users=False)
{<Person: BRANDON MALLORY>: ['add_school']}

[...]

Issue #49, opened 2011-09-03
user.has_perm("perm", obj) behaves unexpectedly


Dzejkob:
If I use standard `user.has_perm("perm")` method, then it will return `True` only, if user has a global permission `"perm"`.
And if `user.has_perm("perm", obj)` is used, it'll teturn `True` if user have permission to access this particular object.
But it will return `False`, even if user has a global permission `"perm"`, which is quite unexpected for me, because I assume, that having global permission should give user access to all objects. Am I right?

xoen:
Hello, I found this issue as I'm trying jango-guardian and I found this behaviour buggy/very confusing.

I added a global permission (let's call it `view_user`) to a user (let's call it `joe`) and then checked if Joe can view a specific user (let's call it `other_user`) and surprisingly it returned `False`.

```python
joe = User.objects.get(username="joe")
other_user = User.objects.get(username="other_user")
assign_perm("myapp.view_user", joe)
joe.has_perm("myapp.view_user") # True as expected
joe.has_perm("myapp.view_user", other_user) # False, whaaaat?
```

Now, I'm not using the `permission_required` decorator explicitly as my view sets are "automatically" checking the permissions because of my `REST_FRAMEWORK/DEFAULT_PERMISSION_CLASSES` and `AUTHENTICATION_BACKENDS` in `settings.py`. How do I tell guardian to behave as expected then?

(Apologies if I'm missing something really stupid, this is kind of day 2/3 in djangoland)

xoen:
Sorry for the noise today, even more interesting/confusing that [`guardian.shortcuts.get_objects_for_user()`](http://django-guardian.readthedocs.io/en/stable/api/guardian.shortcuts.html#std:shortcut-get_objects_for_user) honours the auth/global permissions by default (`accept_global_perms`).

```
accept_global_perms: if True takes global permissions into account. 
[...]
Default is True.
```

doganmeh:
Although, this issue was closed in favor of another, there are a lot of insights here, so I am writing here to re-stir the conversation. Zen of the Python says: 

> There need to be one --and preferably only one-- obvious way to do it.

And to me that is to _fallback to global when local (object level) fails_. The order does not change the result (since Django `returns False if obj is not None`), but may change the performance. 

[...]

PR #546, opened 2018-01-11
Fallback to Model Permissions (Core)

doganmeh:
Related to #380, and #49.

This updates core.py and test_core.py to allow the following checker functions optionally fallback to (or include) model level permissions:

```
get_user_perms(self, obj, fallback_to_model=None)
get_group_perms(self, obj, fallback_to_model=None)
get_perms(self, obj, fallback_to_model=None)
```

Please note the fallback_to_model parameter. 

This also includes, a global setting two that end: `FALLBACK_TO_MODEL`. Function level takes precedence if both provided. Overall defaults to False in order not to change the default behavior. 

When get_perms is used with `fallback_to_model=True`, the combined permissions go through the existing caching mechanism, so a performance degradation is not expected (although not tested). 

Tests are implemented, but documentation is not. More is planned.

### DIFF
--- a/guardian/compat.py
+++ b/guardian/compat.py
@@ -144,10 +144,21 @@ def remote_field(field):
     return field.remote_field
 
 
-def remote_model(field):
+def get_remote_model(field):
     if django.VERSION < (1, 9):
         return remote_field(field).to
     return remote_field(field).model
+
+def get_related_model(field):
+    if django.VERSION < (1, 8):
+        return getattr(field, 'field_model', None)
+    return getattr(field, 'related_model', None)
+
+def get_model_fields (model):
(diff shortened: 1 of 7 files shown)


doganmeh:
previous commit adds standalone cache for model level permissions even if directly accessed via `get_user_perms` and `get_group_perms`

doganmeh:
I now more firmly believe that falling back to model permissions should be via falling back to django default backend, not within Guardian. While this is in the mean time (that was what it was meant to be) useful; should eventually be get rid of. I also have a pull request at Django to help this. Maybe it is better not to pull this in, in hopes that, that one will be.

PR #548, opened 2018-01-12
Update get_users_with_perms() to avoid a large join

aaronmader:
This code was creating a large number of table joins
"Users that are part of a group that have perms to obj" requires that we join 5 tables: user - m2m - group - m2m - obj

Instead; we perform an intermediate query to fetch all of the groups that have a perm record to the object (a join across 3 tables), and then we fetch all of the users that belong to one of those groups (a join across 3 tables).

### DIFF
--- a/guardian/shortcuts.py
+++ b/guardian/shortcuts.py
@@ -32,7 +32,7 @@ def assign_perm(perm, user_or_group, obj=None):
       ``Permission`` instance.
 
     :param user_or_group: instance of ``User``, ``AnonymousUser``, ``Group``,
-      list of ``User`` or ``Group``, or queryset of ``User`` or ``Group``; 
+      list of ``User`` or ``Group``, or queryset of ``User`` or ``Group``;
       passing any other object would raise
       ``guardian.exceptions.NotUserNorGroup`` exception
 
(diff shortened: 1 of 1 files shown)


keattang:
Is there anything else needed for this to be merged? Just had my production database taken down because this query was taking more than 30s

ad-m:
@aaronmader , could you rebase?

aaronmader:
@auvipy I'd be happy to whip something up, but I'll admit that I don't come from a unittesting background. Can you point me to a resource on how I'd write a unittest for a performance gain? Sorry that this is a bit of a naive question.

I assume I can't just set a time-value limit like: "It must run in under 200ms", because there'd be too much variability between different workstations and database backends (sqlite vs mysql).

So what would I be testing? Just that the new code runs faster than the old code?

brianmay:
I think the unit test doesn't need to test the performance gain. Rather it should demonstrate that the function call still works as required (assuming there isn't already such a unit test), e.g. by doing some sample queries for different cases.

aaronmader:
It looks like the get_users_with_perms shortcut method is well tested in [testapp/tests/test_shortcuts.py:GetUsersWithPermsTest](https://github.com/django-guardian/django-guardian/blob/cea980c5ddb1907e891734de45992b15651bf689/guardian/testapp/tests/test_shortcuts.py#L284)

PR #551, opened 2018-01-17
Attaching _obj_perms_cache to user maintains it btw calls to user.has_perm()

doganmeh:
Attached `_obj_perms_cache` to user or group in order to maintain it between successive calls to `user.has_perm(perm, obj)`


thclark:
Forgive me if I'm wrong but I believe an equivalent performance improvement was added subsequent to this PR, so I'm closing to clear up... but thanks for your contribution @doganmeh

PR #555, opened 2018-01-30
Add kwarg to get_users_with_perms to filter for specific permissions as discussed in #246, #383 (rename, add tests)

satyanash:
Fix/Improve the functionality in #246 and #383. Also add tests.

I took code from the PR that @timabbott made and renamed the parameter to `only_with_perms_in` and added a test to check for proper behaviour with the `with_group_users` parameter to `False`.

I want this functionality in the library because I have a few models with multiple permissions so running a `get_users_with_perms` and then manually filtering the result gets old really fast.

@mitar might also be interested.

### DIFF
--- a/guardian/shortcuts.py
+++ b/guardian/shortcuts.py
@@ -212,7 +212,7 @@ def get_perms_for_model(cls):
 
 
 def get_users_with_perms(obj, attach_perms=False, with_superusers=False,
-                         with_group_users=True):
+                         with_group_users=True, only_with_perms=None):
     """
     Returns queryset of all ``User`` objects with *any* object permissions for
     the given ``obj``.
@@ -230,6 +230,10 @@ def get_users_with_perms(obj, attach_perms=False, with_superusers=False,
       **not** contain those users who have only group permissions for given
       ``obj``.
 
+    :param only_with_perms: Default: ``None``. If set to an iterable of
+      permission strings then only users with those permissions would be
+      returned.
+
     Example::
 
         >>> from django.contrib.flatpages.models import FlatPage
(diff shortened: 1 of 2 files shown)


satyanash:
May I get an update on when/if this will be merged?

satyanash:
Can someone review if the code looks okay?

ad-m:
I do not have comments, but let us make a statement to the others who are entitled to approval.

satyanash:
May I get an update on this?

satyanash:
Can someone merge this?

lukaszb:
@satyanash so it looks cool but please provide update to the docs too. I've merged it before realising it so please create another PR

satyanash:
@lukaszb I thought changes to the docstring would've been sufficient. Please let me know where else the documentation needs to be updated?

lukaszb:
Allright, yeah, after second thought, docstring should be enough here. Thanks again!

aaronmader:
Sorry, I'm not sure where best to post my comment.. 

It looks like there's a bug here: 
```
            if only_with_perms_in is not None:
                for permission in only_with_perms_in:
                    group_filters.update({
                            'groups__%s__permission' % group_rel_name: Permission.objects.get(
                                content_type=ctype, codename=permission)
                            })
```

The for-loop here will update the same dictionary key on every iteration, replacing any previous value. 

Also, performance-wise, I'm thinking we should probably fetch all of the Permission records in a single query, and then apply it to group_filters using an __in query.

ad-m:
> The for-loop here will update the same dictionary key on every iteration, replacing any previous value.

Could you write test case for that as proof and as regression protection?

PR #562, opened 2018-04-10
Allow for permission codenames containing `.`s

JamesMaroney:
Limiting the `split()` operation here just like you guys already do in various functions in `guadian.shortcuts`
https://github.com/django-guardian/django-guardian/blob/devel/guardian/shortcuts.py#L78

### DIFF
--- a/guardian/backends.py
+++ b/guardian/backends.py
@@ -82,16 +82,19 @@ class ObjectPermissionBackend(object):
             return False
 
         if '.' in perm:
-            app_label, perm = perm.split('.', 1)
-            if app_label != obj._meta.app_label:
-                # Check the content_type app_label when permission
-                # and obj app labels don't match.
-                ctype = get_content_type(obj)
-                if app_label != ctype.app_label:
-                    raise WrongAppError("Passed perm has app label of '%s' while "
-                                        "given obj has app label '%s' and given obj"
-                                        "content_type has app label '%s'" %
-                                        (app_label, obj._meta.app_label, ctype.app_label))
+            ctype = get_content_type(obj)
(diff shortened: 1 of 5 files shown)


JamesMaroney:
I've added some tests locally - and I don't see an elegant way to support my proposed changes and still keep the `WrongAppError` raised in `backends.py`. 

That code assumes that if there is a dot, then the left side of it is intended to be the `app_label`. If it doesn't match the `app_label` of the provided `obj`, it raises a `WrongAppError`.

If there is any interest in modifying that behavior, I am willing to keep moving forward ... but if it's a dead end let me know. I've already modified my production code to just not use dots in the permissions codenames, so I'm happy either way.

As an alternative suggestion ... I could instead provide a Pull Request that raises a suitable exception when trying to use dots in a codename to keep others from accidentally going down this path too.

michael-k:
> If there is any interest in modifying that behavior, I am willing to keep moving forward ...

2 years later nobody has shown any interest.  Therefore I think we should close the PR.  Sorry.

> I've added some tests locally - and I don't see an elegant way to support my proposed changes and still keep the WrongAppError raised in backends.py.
> 
> That code assumes that if there is a dot, then the left side of it is intended to be the app_label. If it doesn't match the app_label of the provided obj, it raises a WrongAppError.

Maybe if you're using `.`s in your codenames, you alway have to use the `app_label.perm` form or instances of `Permission` and cannot use the `perm` shortcut. :shrug:

Maybe we should document somewhere the using `.`s in permission codenames is not (well) supported.

thclark:
Clsong PR to keep existing functionality. Thanks for the contribution nevertheless!

PR #563, opened 2018-04-11
Allow for UUID ids in admin url routing

JamesMaroney:
This was the only thing keeping django-guardian from completely working with my custom user model using UUID ids.

### DIFF
--- a/guardian/admin.py
+++ b/guardian/admin.py
@@ -102,11 +102,11 @@ class GuardedModelAdminMixin(object):
                     view=self.admin_site.admin_view(
                         self.obj_perms_manage_view),
                     name='%s_%s_permissions' % info),
-                url(r'^(?P<object_pk>.+)/permissions/user-manage/(?P<user_id>\-?\d+)/$',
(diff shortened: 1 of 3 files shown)


JamesMaroney:
I tried a handful of approaches at writing a test for this new relaxed behavior, foiled at every step - so I am open to suggestions on how you'd go about swapping in a new user (and group) model for just one or a couple of tests.

Obviously I can't change `guardian.testapp.models.CustomUser` because tests rely on it's id being an int. Introducing a second subclass of `AbstractUser` confuses everything. Subclassing `AbstractBaseUser` seems the right way to go, but after following the bouncing ball of exceptions for a while, it seems I was ultimately missing something as it was still trying to lookup against `CustomUser`.

Any hints you can provide would be very much appreciated. Thanks!

JamesMaroney:
This last commit adds in a broken (skipped) test that is the closest to what I think could potentially work. I'm hoping it is close enough that someone can let me know how to get it across the finish line.

bldm95:
@JamesMaroney You forgot to create a migration

thclark:
If anyone wants to follow up and fix this, completion of the outstanding PR (make the tests work) would be welcome!

PR #566, opened 2018-04-30
Add missing css class to the Grappelli template

tdruez:
`grp-table` class is now required on HTML <table> with latest Grappelli versions


ad-m:
Thanks!

PR #572, opened 2018-06-27
Allow for adding permission to multiple users or groups

intrepidev:
This allows for using the `assign_perm` shortcut to grant a set of users or groups permissions to an object in bulk.

```
article = Article.objects.get(pk=1)

user_queryset = User.objects.filter(some_attr=some_value).all()
assign_perm("change_article", user_queryset, article)

user_list = [user1, user2, user3, user4]
assign_perm("change_article", user_list, article)

group_queryset = Group.objects.filter(some_attr=some_value).all()
assign_perm("change_article", group_queryset, article)

group_list = [group1, group2, group3, group4]
assign_perm("change_article", group_list, article)
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -56,3 +56,4 @@ Authors ordered by first contribution
 - Frank Wickström <frank@bambuser.com>
 - George Karakostas <gckarakostas@gmail.com>
 - Adam Dobrawy <guardian@jawnosc.tk>
+- Jeff Hackshaw <intrepidevio@gmail.com>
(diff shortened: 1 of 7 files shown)


nightsnow0918:
I have ran into an integrity error (Duplicate entry) when assigning perm to user/group queryset that some of them has already been assigned that perm. For example, there is a queryset contains `group1`~`group4`. Yet `group1` has already been has assigned `perm1` on an object. Then if I do `assign_perm('perm1`, queryset, object)`, it will raise the integrity error.

Maybe a `has_perm` check before doing `bulk_create()` like what has done in [`bulk_assign_perm()`](https://github.com/django-guardian/django-guardian/blob/devel/guardian/managers.py#L67-L76) is necessary?

Sorry for replying here since I am not sure if this should be opened with a new issue. If this is not appropriate, plz let me know.

PR #578, opened 2018-07-27
#577 Add translations for simplified Chinese

### LINKED ISSUE #577: Add translations for simplified Chinese
I will create a pull request soon.

Jay54520:
According to http://www.i18nguy.com/unicode/language-identifiers.html , the language code of simplified Chinese is "zh-Hans", so the locale name is "zh_Hans".

Fix #577


Jay54520:
Where is the position of "操作"? Why don't you comment in "Files changed"?

ad-m:
@lintingbin2009 (author #591), could you review that pull request? I do not know Chinese, so it is difficult to judge whether pull request is correct, if users do not discuss and do not accept.

lintingbin2009:
@ad-m This PR is good, I think you should merge it.

Issue #582, opened 2018-09-11
Python 3.7 and Django 2.1 compat


michael-k:
I'm working on this.  If anyone else want to help, please comment on this issue for coordination and to avoid duplicate work.

hmpf:
Any chance of a new version that supports 3.7? I see that the code is in, but there is no new version since 1.4.9, and that one's two years old.

PR #584, opened 2018-09-11
Add support for Django 2.1+

michael-k:
This includes the commit from django-guardian/django-guardian#583 because otherwise testing would not make much point.  I'll rebase as soon as django-guardian/django-guardian#583 is merged.

Regards django-guardian/django-guardian#582


michael-k:
@ad-m rebased

PR #593, opened 2018-11-02
enhance admin integration

lintingbin:
1. only superuser to view the "Object permissions" button.
2. overwrite has_module_permission, get_queryset, has_view_permission, has_change_permission, has_delete_permission function of the GuardedModelAdmin class.
3. auto add obj permission to the user who create it. 
4. auto delete obj permission after obj be deleted.


guardian/admin.py
```diff
@@ -428,6 +428,66 @@ class AuthorAdmin(GuardedModelAdmin):
...
+
+    @staticmethod
+    def remove_obj_perms(obj):
+        perms_dict = get_users_with_perms(obj, attach_perms=True)
+        perms_dict.update(get_groups_with_perms(obj, attach_perms=True))
+        for user_or_group in perms_dict:
+            [remove_perm(perm, user_or_group, obj) for perm in perms_dict[user_or_group]]
+
+    def delete_model(self, request, obj):
+        self.remove_obj_perms(obj)
```
ad-m: In the case of Generic Foreign Key, such an operation is useful (as far as I know). However, in the case of Direct Foreign Key, it is responsible for the database, so there is no point in performing this operation.

I am afraid, however, why this operation is to be performed only when using the administration panel. Maybe we should register the listeners of the corresponding signals?

guardian/admin.py
```diff
@@ -428,6 +428,66 @@ class AuthorAdmin(GuardedModelAdmin):
...
+            [assign_perm('{}.{}_{}'.format(opts.app_label, action, opts.model_name), request.user, obj)
+             for action in actions]
+        return result
+
+    @staticmethod
+    def remove_obj_perms(obj):
+        perms_dict = get_users_with_perms(obj, attach_perms=True)
+        perms_dict.update(get_groups_with_perms(obj, attach_perms=True))
+        for user_or_group in perms_dict:
+            [remove_perm(perm, user_or_group, obj) for perm in perms_dict[user_or_group]]
```
ad-m: We extract the code for "shortcuts" instead of static methods. This allows you to re-use the code.

You download the permissions (+1 query) to then delete the permission for each user (+n queries). Could you eliminate this and build the correct query that will do it all at once? Maybe we need to extend our sho

guardian/admin.py
```diff
@@ -428,6 +428,66 @@ class AuthorAdmin(GuardedModelAdmin):
...
+
+    def has_delete_permission(self, request, obj=None):
+        return self.has_perm(request, obj, 'delete')
+
+    def save_model(self, request, obj, form, change):
+        result = super(GuardedModelAdmin, self).save_model(request, obj, form, change)
+        if not request.user.is_superuser and not change:
+            opts = self.opts
+            actions = ['view', 'add', 'change', 'delete']
+            [assign_perm('{}.{}_{}'.format(opts.app_label, action, opts.model_name), request.user, obj)
```
ad-m: This is transferring the internal logic of your application to the library, which should be deprived of such elements. Why do not you use signals for this?

guardian/admin.py
```diff
@@ -428,6 +428,66 @@ class AuthorAdmin(GuardedModelAdmin):
...
+        return get_objects_for_user(user=request.user, perms=['{}_{}'.format(perm, model_name) for perm in actions],
+                                    klass=klass, any_perm=True)
+
+    def has_module_permission(self, request):
+        if super(GuardedModelAdmin, self).has_module_permission(request):
+            return True
+        return self.get_model_objs(request).exists()
+
+    def get_queryset(self, request):
+        if request.user.is_superuser:
```
ad-m: Again, moving  logic of your application to our library. Is there a shortcut to this?

guardian/templates/admin/guardian/model/change_form.html
```diff
@@ -2,7 +2,9 @@
 {% load i18n admin_urls %}
 
 {% block object-tools-items %}
+{% if request.user.is_superuser %}
```
ad-m: I can not accept that way to determine button visibility. It introduces a security risk because the button will have a different display logic from the page to which it directs. I agree that this should be uniform and this is currently an bug. However, we can not solve this in such a simplified way, because we lose flexibility. I rather see in that part reusing that condition: https://github.com/django-guardian/django-guardian/blob/devel/guardian/admin.py#L225
It is possible that this condition should be extracted to separate method, so the result of that method can be overwritten (extra flexibility for end user) and used in two places.

guardian/admin.py
ad-m: I understand that you need information about who created it. Usually, I stores such information in the object (createdBy field), so the signal listener have access to it.

guardian/admin.py
lintingbin2009: When an obj is deleted, we need have an api to delete its permission. I think this api put to "shortcuts" is good.

guardian/admin.py
lintingbin2009: I think this isn't the internal logic of my application, this function is the normal function that django-guardian need to provide when django-guardian be integrated with Admin.

ad-m: The term "internal logic" was not the best term. I see the field for improvement and after thinking it out I believe it is valuable to give permission to the user who created the resource. In my opinion, a persistent set of permissions is not suitable for library. This functionality may work, for example, without using "delete" permissions. This set of permissions can be the default.

guardian/admin.py
ad-m: I get the impression that transferring logic, because in django-guardian, if there is a possibility, we do not abuse the superuser, and in fact we use permissions. Independently, I see the potential for change in this.

guardian/admin.py
ad-m: Could you update pull request to reflect that?

guardian/admin.py
lintingbin2009: I think this is no problem. The superuser have the all permission, and "get_queryset" return all objs for superuser is correct.

thclark:
Im closing this because there are change requests outstanding after 6 years. Also, it seems to me this would be a breaking change since it fundamentally changes what permissions are set on objects; a new workflow/user training would be required. The use case isn't clear because the admin is used in very diverse ways; this assignment of permission presupposes particular usage patterns.

It really seems like signals are the right way to do this.

PR #599, opened 2019-01-25
Add admin support for user primary key as uuid

lukewiwa:
I noticed this route would error out since the regex only looks for digits. Here's a fix.


guardian/admin.py
```diff
@@ -102,11 +102,11 @@ def get_urls(self):
                     view=self.admin_site.admin_view(
                         self.obj_perms_manage_view),
                     name='%s_%s_permissions' % info),
-                url(r'^(?P<object_pk>.+)/permissions/user-manage/(?P<user_id>\-?\d+)/$',
+                url(r'^(?P<object_pk>.+)/permissions/user-manage/(?P<user_id>.+)/$',
                     view=self.admin_site.admin_view(
                         self.obj_perms_manage_user_view),
                     name='%s_%s_permissions_manage_user' % info),
-                url(r'^(?P<object_pk>.+)/permissions/group-manage/(?P<group_id>\-?\d+)/$',
+                url(r'^(?P<object_pk>.+)/permissions/group-manage/(?P<group_id>.+)/$',
```
jpic: why not make the move to path while were at it ?

ad-m:
PR is not possible to merge because it contains everything. @lukewiwa , can you rebase? Users in issue #611 awaiting for that PR

lukewiwa:
Looks like this has been resolved in the latest devel branch by changing the urls to paths anyway. Should I still rebase then? Thinking it might not be worth it.

ad-m:
@lukewiwa , 10 hour ago I implemented that by self (see #634). 6 hour ago I released new version which includes that changes. Any feedback welcome!

ad-m:
BTW. Do you have still any needs to keep own fork?

lukewiwa:
@ad-m nah will delete

PR #605, opened 2019-02-22
Call `string.split` without kwargs for Python 2.7 compat

michael-k:
Releasing version 1.5.1 with a fix for Python 2.7 support would be the polite thing to do.  This PR paves the way towards that.

Fixes django-guardian/django-guardian#602


AGASS007:
hi guys,

what's preventing us from merging?

cheers

AGASS007:
\o/

PR #614, opened 2019-06-26
Add database index to object_pk field in BaseGenericObjectPermission

dblenkus:
`object_pk` field is used for filtering (eg [here](https://github.com/django-guardian/django-guardian/blob/ab4e14e76c9228a7f6e256399799e1439c441b85/guardian/shortcuts.py#L275), [here](https://github.com/django-guardian/django-guardian/blob/ab4e14e76c9228a7f6e256399799e1439c441b85/guardian/shortcuts.py#L291) and [here](https://github.com/django-guardian/django-guardian/blob/ab4e14e76c9228a7f6e256399799e1439c441b85/guardian/shortcuts.py#L367)), so a database index can significantly speeds up those queries, especially on big production databases.


ad-m:
Why not combined index (```index_together```) on ```object_id``` and ```content_type```?

dblenkus:
Good point! I've updated the pull request.

ad-m:
Currently a supported version of Django is 2.1+. Since 1.11 ```index_together``` is deprecated, use ```indexes``` instead to avoid introducing technical debt. See Django docs https://docs.djangoproject.com/en/2.2/ref/models/options/#index-together also.

dblenkus:
Thanks for pointing this out. Fixed.

ad-m:
Thank you for your cooperation and project development.

PR #623, opened 2019-07-15
Update example to be PEP8 compliant

johnthagen:


johnthagen:
@ad-m You merged the other PR so quickly, that this PR looked like it tried to revert it 🤣.

Feel free to squash this PR when you merge into a single commit.

ad-m:
I think it would be a good idea to shape the example project and insert the right comments so that examples in the documentation can be obtained using ```literalinclude```. See for works of @mik-laj in Apache Airflow ( https://airflow.readthedocs.io/en/latest/howto/operator/bash.html , https://github.com/apache/airflow/blob/master/docs/howto/operator/bash.rst ). It allows us to automatically tests & lint examples of codes.

johnthagen:
Could that be a separate PR?

ad-m:
Yes of course. I share the idea, because I noticed that you want to develop the documentation of this project. The project does not have large human resources, so I think how to avoid errors and to reduce human effort to maintain.

Issue #630, opened 2019-08-19
post_migrate signal always uses the default database


AdamZ:
In the `create_anonymous_user` function, the user lookup and save calls don't specify the `using` parameter/method call. This causes the operations to always apply to the default database, even if the migration is occurring on a different database.

It [looks like](https://docs.djangoproject.com/en/2.2/ref/signals/#post-migrate) the post_migrate signal sends a `using` parameter to the callback to tell it which DB is used, so it should be simple to add it to the `get()` and `save()` calls in the signal handler.

ad-m:
Could you provide PR for that?

ad-m:
Could you close that issue?

PR #631, opened 2019-08-19
Use the correct DB in post_migration handler

AdamZ:
Django provides a keyword argument called `using` to the `post_migrate`
signal handler. This argument tells the handler which database was
migrated.

Previously, the `create_anonymous_user()` function would always try to
use the default database to create the anonymous user. This change uses
the `using` parameter to make sure we use the database that actually
underwent the migration.

Closes: django-guardian/django-guardian#630

### DIFF
--- a/guardian/management/__init__.py
+++ b/guardian/management/__init__.py
@@ -28,12 +28,12 @@ def create_anonymous_user(sender, **kwargs):
     User = get_user_model()
     try:
         lookup = {User.USERNAME_FIELD: guardian_settings.ANONYMOUS_USER_NAME}
-        User.objects.get(**lookup)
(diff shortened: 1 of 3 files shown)


AdamZ:
I've checked the Django code and as far as I can tell it always provides the `using` argument to the signal. I've also verified that this change fixes the issue I was having in my own code (caused by a blank `default` database config).

ad-m:
Is there any generic way to identify other bugs like that?

AdamZ:
Not as far as I know. Basically any query or `save()` call can specify which database to use. Generally, the default DB is used most of the time, but any time the user is doing something specific for one database you might need to specify the DB. I don't know enough about the django-guardian internals to know when it's appropriate to specify the DB config to use.

PR #632, opened 2019-08-27
Auto prefetching for vanilla usecases

qeternity:
I threw this together for our own internal use which involves iterating over many various object types through many libraries and interfaces. It simply wasn't feasible to do explicit prefetching the way that django-guardian currently allows. This expects a pretty vanilla setup so definitely not for everyone, but would be nice to have some out-of-the-box functionality that does this. A query-per-check architecture is too onerous for some people.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -153,3 +153,17 @@ polymorphic models and the regular model ``ContentType`` for non-polymorphic
 classes.
 
 Defaults to ``"guardian.ctypes.get_default_content_type"``.
+
+GUARDIAN_AUTO_PREFETCH
+-------------------------
+
+.. versionadded:: 2.x.x
+
+For vanilla deployments using standard ``ContentType`` interfaces and default
+``UserObjectPermission`` or ``GroupObjectPermission`` models, Guardian can automatically
+prefetch all User permissions for all object types. This can be useful when manual prefetching
+is not feasible due to a large number of model types resulting in O(n) queries. This setting may
+not be compatible with non-standard deployments, and should only be used when non-prefetched
+invocations would result in a large number of queries or when latency is particularly important.
+
+Defaults to ``False``.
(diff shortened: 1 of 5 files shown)


ad-m:
Could you add the documentation indicating the use case of new feature, the benefits of the solution and its limitations, when its use is recommended and when it should definitely not be used?

qeternity:
Added to configuration docs, although would deserve a mention in performance tuning if merged.

The need for this patch internally is that we check object permissions across many different contenttypes in many different locations in the code. It's not always possible to pass a prefetched permissions checker object into these functions, and even if we were able to, each prefetch requires its own query so for N contenttypes you have N queries. The above patch resolves all permissions available to a user in 2 queries (one for groups, one for users).

Per the docs this should not be used in non-standard deployments (our setup is fairly vanilla when it comes to perms and contenttypes) and as with any prefetching, it should not be used when the ratio of total number of rules (objects * permissions) is orders of magnitude greater than the number of saved queries (i.e. it does not make sense to prefetch 1mm permissions to save 5 queries).

ad-m:
There is still a lack of testing to ensure that the code will work with future versions of Django.

qeternity:
I can add some tests to verify that the caching is actually working, but the only Django compatibility issue would be namespace on the User model (i.e. being able to setattr and delattr _guardian_perms_cache)

Do you think testing the above two conditions would suffice? Or can you think of a better way to test Django compatibility?

qeternity:
I've added some tests, as well as branches to ensure this only activates against User objects. I'm happy to add functionality for Groups as well (not something we do in current project) if this PR is something worth merging

ad-m:
I think that this is an element worth incorporating into the project, which is why I am cooperate to bring the change to a level acceptable by the project.

qeternity:
I added Group support and some tests. All of the tests mimic the existing prefetch tests except they add a few more checks.

ad-m:
Looks good to me. Do you want to add anything else, can I merge it?

qeternity:
@ad-m fire at will Commander

PR #637, opened 2019-09-09
Improve performance of get_objects_for_user, removing sets

woakas:
Hi,

Since django 1.8 it is not necessary to use a "set" or a list it can be directly replaced inside the id__in parameter, the queryset will always receive the same type of content type and the validation is not necessary.


ad-m:
@woakas, le'ts look #366 and CI tests results, especially for psql.

woakas:
@ad-m Done, I fixed the test, let me know your thoughts about this PR.

PR #642, opened 2019-09-24
adding django3.0

samims:


tox.ini
```diff
@@ -33,3 +35,4 @@ deps =
     docs: setuptools_scm
     django21: django>=2.1,<2.2
     django22: django>=2.2,<2.3
+    django30: django>=3.0a1,<3.1
```
ad-m: Use ```django~=3.0```  instead.

ad-m:
It looks that tests fail. Non-mergable in that state.

samims:
What should I do now?

tox.ini
ad-m: It seems that this is not correct until version 3.0.0 is released. Let him stay that way for now.

ad-m:
Making all necessary changes to make the tests pass. See at https://travis-ci.org/django-guardian/django-guardian/builds/588877851 for logs.

tox.ini
```diff
@@ -3,6 +3,7 @@ downloadcache = {toxworkdir}/cache/
 envlist = # sort by django version, next by python version
     {core,example,docs}-py{35,36,py37}-django21,
     {core,example,docs}-py{35,36,py37}-django22,
+    {core,example,docs}-py{35,36,py37}-django30,
```
ad-m: > Django 3.0 supports Python 3.6, 3.7, and 3.8. We highly recommend and only officially support the latest release of each series.
> The Django 2.2.x series is the last to support Python 3.5.

.travis.yml
```diff
@@ -15,6 +15,7 @@ notifications:
 templates:
   django21: &django21 DJANGO_VERSION=2.1.*
   django22: &django22 DJANGO_VERSION=2.2.*
+  django30: &django30 DJANGO_VERSION=3.0.*
```
auvipy: ERROR: No matching distribution found for Django==3.0.*, could you try 3.0a1 instead?

tox.ini
michael-k: `django~=3.0` won't be correct after the release of 3.0.0 as well.  It's equivalent to `>=3.0, ==3.*`.  https://www.python.org/dev/peps/pep-0440/#compatible-release

`django~=3.0.0` is what you were looking for. :)

ad-m:
Closed by #667

Issue #423, opened 2016-04-25
Django should not be in setup.py's install_requires


crccheck:
If someone is installing django-guardian, it's safe to assume they have Django installed already. Putting Django in `install_requires` ends up causing more problem than it solves because of unexpected Django upgrades.

Similar discussion: https://github.com/caioariede/django-location-field/issues/47

brianmay:
Sounds like this is different from the django-location field case: "Django isn't a dependency of location-field in any useful sense."  Django is a hard requirement for django-guardian.

Furthermore setuptools should not be touching Django if it already meets the requirement (Django >= 1.7). This is what happens with my tests.

Maybe you have an old/buggy version of setuptools? Or could it be that another package was getting installed at the same time that pulled in a specific version of Django?

Also see #420.

brianmay:
No response; closing. Can reopen if required.

andybak:
Hi,

For me the issue is specifically this: https://github.com/caioariede/django-location-field/issues/47#issuecomment-192703189

Quite often I want to upgrade a requirement locally before I update requirements (and without wanting to do a full requirements update which wipes out any local changes to -e installed packages) - it's surprising if doing this also upgrades Django itself. (of course it's possible that other packages get updated but this is both less common and less unpleasant to deal with)

The other side of this - what benefit is there in including Django in install_requires which isn't better served by using the system check framework?

Apologies if I've misunderstood something.

brianmay:
The only requirement django-guardian has is:

```
Django >= 1.7
```

pip not not be trying to upgrade Django unless it is too old for django-guardian to support.

crccheck:
Frequently, pip will see that and "helpfully" upgrade Django versions for you when you already have a recent Django version.

ad-m:
@brianmay , I disagree. pip trying to upgrade django everytime upgrade django-guardian.

```
$ pip install 'django==1.8'
Collecting django==1.8
  Using cached Django-1.8-py2.py3-none-any.whl
Installing collected packages: django
Successfully installed django-1.8
$ pip install django-guardian
Collecting django-guardian
Requirement already satisfied (use --upgrade to upgrade): six in ./test-django/lib/python2.7/site-packages (from django-guardian)
Requirement already satisfied (use --upgrade to upgrade): Django>=1.7 in ./test-django/lib/python2.7/site-packages (from django-guardian)
Installing collected packages: django-guardian
Successfully installed django-guardian-1.4.5
[...]

PR #646, opened 2019-09-30
Add requirements (specifically, Django >= 2.1) to setup.py

kdmccormick:
[Python's official documentation](https://packaging.python.org/discussions/install-requires-vs-requirements/) describes that packages should declare their installation dependencies in setup.py in order to play nicely with setuptools.

I enjoy using `django-guardian` in one of my projects, but I currently have to handle it differently than all of my other dependencies because it doesn't use `install_requires`.

I'm not sure whether `bumpversion` needs to be in `install_requires`. Let me know what you think.

Thanks!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
     ],
     include_package_data=True,
     license='BSD',
-    install_requires=[],
+    install_requires=["Django>=2.1"],
     tests_require=['mock', 'django-environ', 'pytest', 'pytest-django'],
     classifiers=['Development Status :: 5 - Production/Stable',
                  'Environment :: Web Environment',


ad-m:
Bumpversion is not requried to install django-guardian. It's only for development management purpose like eg. CI configuration.

Django was removed from ```install_requires``` ( https://github.com/django-guardian/django-guardian/commit/2eefc7b4c325028d42b28a4d52d5c6ac6a4891e3#diff-2eeaed663bd0d25b7e608891384b7298 ) for purpose. See  #423, #710 .

kdmccormick:
@ad-m I pushed a new commit that doesn't include `bumpversion`.

Respectfully, I disagree with the conclusion of #423. The point of putting Django in setup.py isn't to transitively install Django... of course everyone using Django Guardian already requires Django. The point is to codify the Django version constraint (>=2.1, currently) so that package managers like pip and upgrade managers like [pip-tools](https://github.com/jazzband/pip-tools/) can correctly select a version of Django Guardian version of install/pin that is compatible with the installed/pinned Django version.

Our project (https://github.com/edx/registrar) uses `pip-tools`, and our upgrade pipeline was thwarted when it selected Django Guardian 2.1, which doesn't support our current Django version (1.11). None of the other packages that our project requires have this issue.

This request is not specific to my use case; it is specified in the official Python documentation, and is widely used by most packages. _Not_ wanting to have requirements defined in `setup.py` is the special case.

PR #653, opened 2019-11-06
Allow custom generic object permission models

partimer:
Override id from 32bit Integer to 64bit Integer.  Raises the limit of total object permissions to 2^63


ad-m:
How do other Django projects usually approach the "ID" field?

ad-m:
Missing migration...

partimer:
id AutoField is automatically created by django Model -- if it is not overridden

So this is likely the most typical for many projects.

Keeping under 2^31 aggregate lifetime instantiations per model is pretty typical.  However,  UPO and GPO are potential aggregates of other model instantiations. Thus it becomes less likely to stay comfortably away from 2^31.

brianmay:
It looks like this change will require applications to be updated also, when they have foreign keys that point to this ID. This includes the demo app, included with Guardian. As per travis logs:

```
Migrations for 'articles':
  articles/migrations/0002_auto_20191106_1221.py
    - Alter field id on articlegroupobjectpermission
    - Alter field id on articleuserobjectpermission
The command "./contrib/travis/test.sh" exited with 1.
```

partimer:
Is the Travis failure because it auto generated a migration?

ad-m:
We have tests for missing migration in TravisCI. Any migration generated on TravisCI make tests fail as missing migration in commit.

ad-m:
See https://github.com/django-guardian/django-guardian/blob/devel/contrib/travis/test.sh#L12 .

ad-m:
I wonder if we should make such a change in this way. This will affect all projects that use our library. The migration process itself will require considerable attention from users. I am not sure how this change will affect a number of database engines that we support, because this is not the default behavior of Django (probably for a reason).

brianmay:
I would hope that all databases support BigAutoField. Although I can't be sure of this. It could be legacy reasons however why it isn't the default. Changing the default would break every deployed Django application.

I am not fond of the way Django encourages libraries (such as Guardian) to come with their own database tables (yes, I use to like this). It means if you want to make any changes to the supplied tables for your own application (no matter how simple or complicated), you have limited options,  all of them generally bad:

* Get change in upstream library.
* Fork upstream library.
* Add config options in upstream library.
* Add ability to substitute db model class for another class (similar to Django user).
* Monkey patch upstream library (no idea if this is possible; don't want to think about it)
* In this case, manually adjusting the database outside migrations might work.
* In other cases adding a one-to-one relationship with another table can help add extra fields.
* Probably others I can't think of right now.

(as an example of another approach, there is a Guardian library for Elixir, that does token based authorization, but instead of providing its own database tables it requires 2 hooks (functions) that retrieves the data from the database on its behalf)

Another solution here might be to use BigAutoField if and only if there is some setting, e.g. settings.GUARDIAN_USE_BIG_ID that is defined. Not sure I really like adding even more settings, but something to thing about. It should be simple to implement, and preserves backward compatibility with existing applications.

ad-m:
@partimer, are you use direct foreign key?

partimer:
Not at the moment.  And may not work (well) for particular cases where Models are dependent on permission of GenericKey reference where the permission is compared against the content_type_id+object_pk in both tables.

Having a configurable GUARDIAN_USE_BIG_ID is probably the safest approach.

ad-m:
How to manage migrations when having a configurable GUARDIAN_USE_BIG_ID?

partimer:
If GUARDIAN_USE_BIG_ID=True, then adding a migration to guardian later would likely require "./manage.py makemigrations --merge".

Definitely not clean or smooth....

Is it feasible to switch to a different model similar to get_user_model()/AUTH_USER_MODEL?
i.e. BigUserObjectPermission or BigGroupObjectPermission
- This type approach would avoid the de-serialized migrations.
- Potentially allow overriding of the ObjectPermission Model

What potential pitfalls am I not seeing with this approach?

brianmay:
I am a bit nervous about using something like AUTH_USER_MODEL, but it probably is an option worth exploring all the same. Need to make sure that the migrations under Guardian do not change even if the model changes. It is OK for the applications migrations to change, not the Guardian migrations. I can't remember how this works for `AUTH_USER_MODEL` right now.

partimer:
`AUTH_USER_MODEL` is only automatically supported (and encouraged) when set at the start of a project.  There is an explicit not supported disclaimer after the start of a project.
https://docs.djangoproject.com/en/2.2/topics/auth/customizing/#auth-custom-user

Then internally or externally get_user_model() is used to obtain the project's current User model. 
```
from django.contrib.auth import get_user_model
User = get_user_model()
User.objects.all()
```

Instead of the current:
```
from guardian.models import UserObjectPermission, GroupObjectPermission
```
Something like:
```
from guardian.models import get_group_obj_perm_model,  get_user_obj_perm_model
UserObjectPermission = get_user_obj_perm_model()
GroupObjectPermission = get_group_obj_perm_model()
```
It could also be mandated sub-classes must be derived from the guardian obj perm classes: For example in 'myapp.models':
```
class BigUserObjectPermission(UserObjectPermission):
    id = models.BigAutoField(editable=False, unique=True, primary_key=True)
    class Meta(UserObjectPermission.Meta):
        pass

class BigGroupObjectPermission(GroupObjectPermission):
    id = models.BigAutoField(editable=False, unique=True, primary_key=True)
    class Meta(GroupObjectPermission.Meta):
        pass
```
In settings:
```
GUARDIAN_USER_OBJ_PERM_USE = 'myapp.BigUserObjectPermission'
GUARDIAN_GROUP_OBJ_PERM_USE = 'myapp.BigGroupObjectPermission'
```

-No migration files in guardian (only in 'myapp')
-Downstream libraries not affected (can still just directly access)
-Downstream libraries that do support would just call `get_user_obj_perm_model() `and `get_group_obj_perm_model()`
- Customized Object Permission Models (to a reasonable extent)

_Any caveats I am missing with this approach?_

ad-m:
It is worth noting that in the case of django-guardian we have two types of models available:
- built-in
- external, used for DFK.

Definitely introducing the ability to change the built-in model requires careful consideration & testing. The PR presented here modifies both of these models due to the default class hierarchy. I am not sure if the currently proposed change will work properly for full django-guardian public API, if the configuration option is enabled and the built-in model is used.

In the case of the built-in model, we manage migration. For the external model, the external application manages migrations.  

Regardless of the method of change currently proposed, I see great potential in the way ```AUTH_USER_MODEL``` works:

> ```AUTH_USER_MODEL``` is only automatically supported (and encouraged) when set at the start of a project.

The authors of Django have adopted a reasonable limitation that certain configuration options must be set at the beginning of the project and cannot be easily changed in the future. For django-guardian, it is an acceptable limit for me so that settings ```GUARDIAN_USE_BIG_ID``` can also be set at the beginning of the project.

Therefore, we must now consider how to provide the model and relevant migrations, only at the beginning of the project's life. We must consider each approach through the prism of both types of models that django-guardian currently supports.

We can introduce another way to inject an external model, which is currently proposed. However, I wonder if it is safer (in the sense of simplicity of maintenance) only to introduce a field in build-in model depending on the configuration option (which was proposed at some stage), and thus create a migration that will be sensitive to this configuration option (not proposed at any stage yet).

ad-m:
In other words:

>  Even if we add a migration and make it do nothing if the GUARDIAN_USE_BIG_ID is not set, this will break if somebody enables the setting after running the migration. 

I think this is not the case that we must support, at least not automatically. In this case for ```AUTH_USER_MODEL```, Django documentation refers to manual changes to the scheme, which is acceptable to me for ```GUARDIAN_USE_BIG_ID``` as a very edge case. We can resolve this case through documentation too.

partimer:
My current understanding is `GUARDIAN_USE_BIG_ID` cannot satisfy automatic/smooth migrations (future compatibility).  I would consider this approach dead.

The current approach `GUARDIAN_USER_OBJ_PERMS_USE` and `GUARDIAN_GROUP_OBJ_PERMS_USE` can satisfy automatic/smooth migrations.  This is effectively the same approach as `AUTH_USER_MODEL`  in django.

- Maintains backwards compatibility with existing libraries (backwards compatible)
- Keeps migration changes with the user's application (future compatible)
- Allows users to provided their own generic object permissions (solves user specific cases)

But this approach is far more invasive and effectively adds a new feature.

ad-m:
How approach ```GUARDIAN_USER_OBJ_PERMS_USE``` and ```GUARDIAN_GROUP_OBJ_PERMS_USE``` can satisfy automatic/smooth migrations when ```AUTH_USER_MODEL``` in django cannot satisfy automatic/smooth migrations and approach ```GUARDIAN_USER_OBJ_PERMS_USE``` and ```GUARDIAN_GROUP_OBJ_PERMS_USE```  is effectively the same approach as ```AUTH_USER_MODEL``` in django?

partimer:
`AUTH_USER_MODEL` does have automatic/smooth migrations _when set at the start of a project_.
1. Data is preserved
2. django automatically creates migrations and manual merges are not necessary (i.e. changes to the custom user model)

This is because `AUTH_USER_MODEL` must be set prior to the start of a project (no data to preserve/convert)

When changes to the custom user model are made that generate migrations, these migrations exist in the application containing the custom user model.  Since the custom user migrations are serialized within that application, there is no migration dependency for django.contrib.auth.model.User.


Under the same conditions `GUARDIAN_USER_OBJ_PERMS_USE` and `GUARDIAN_GROUP_OBJ_PERMS_USE` is expected to behave similarly:
- Only set at the beginning of a project (No initial data migration/conversion)
- Custom generic object permission migrations are contained in the user's application, not guardian
- Guardian's object permission migrations are not altered

As far the dynamic determination of the Model `get_user_model()` underneath uses 
`
django_apps.get_model('app_label.ModelName')
`
This requires django applications using a custom user model to call ' User = get_user_model()' to obtain the correct model.

Guardian is 50% of the way there in `guardian.utils` `get_user_obj_perms_model()` and `get_group_obj_perms_model()` when dealing with Direct Foreign Key object permissions.

To get the remaining 50% guardian needs to call `guardian.utils` `get_user_obj_perms_model()` and `get_group_obj_perms_model()` to determine the generic object permission model. (Note: internal to `guardian.utils` `django_apps.get_model()` would need to be called)

Any dependent library that wants to support custom generic object permissions would also need to `guardian.utils` `get_user_obj_perms_model()` and `get_group_obj_perms_model()`.

brianmay:
Who is the change targeted for? Do we expect new projects to benefit or existing projects? I would have existing projects was prime target audience - that is people who have suddenly discovered that the existing id field is not big enough for their needs.

Unfortunately it is existing projects that are hardest to deal with if we attempt to make this change optional.

partimer:
Not sure I follow.  I needed to alter the id field to a BigSerial and additionally add an additional indexes to speed up some SQL queries (but was doing that via direct SQL).  However as @ad-m pointed out, altering the guardian generic object permission models broke future compatibility.

As for the other guardian users, if they have a case where they would need a custom functionality from generic object permissions, `GUARDIAN_USER_OBJ_PERMS_USE` and `GUARDIAN_GROUP_OBJ_PERMS_USE` approach permits it.

partimer:
@brianmay Change [36f3bfa](https://github.com/django-guardian/django-guardian/pull/653/commits/36f3bfa0b09bd00249fa51bef405a923398a4faa) enables custom generic object permission models for all cases.  So, `from guardian.models import UserObjectPermission, GroupObjectPermission` now works dynamically and shouldn't.  If `GUARDIAN_USER_OBJ_PERMS_USE` or `GUARDIAN_GROUP_OBJ_PERMS_USE` are set, then the new model is picked up automatically.
If not set, nothing changes and default generic permission models are used.


@ad-m If `GUARDIAN_USER_OBJ_PERMS_USE` or `GUARDIAN_GROUP_OBJ_PERMS_USE` were set after project start, the underlying table data would have to be manually migrated. For example
```
INSERT INTO articles_biguserobjectpermission (object_pk, content_type_id, user_id)
SELECT object_pk, content_type_id, user_id
FROM guardian_userobjectpermission;
```

- [X] Maintains backwards compatibility with existing libraries (backwards compatible)
- [X] Keeps migration changes with the user's application (future compatible)
- [X] Allows users to provided their own generic object permissions (solves user specific cases)
- [X] Public API unchanged, feature enabled for existing libraries (universally enabled)

_What else should be added to the check list?_

partimer:
Closing for #657 

Created another branch that only has `GUARDIAN_USER_OBJ_PERMS_USE` and `GUARDIAN_GROUP_OBJ_PERMS_USE` related changes

This way the request changes can be more clearly seen

PR #655, opened 2019-11-09
Upgrade syntax to Python 3 

ad-m:


ad-m:
I am aware that this change is incompatible with Python 2.7. In #604 we drop support for Python 2.7. This PR is continuation of clean up of that PR.

I prepared it because I discovered tool ```pyupgrade``` that allows us to easily use the new syntax and simplify our code.

PR #657, opened 2019-11-14
Feature: Custom generic object permission models

partimer:
Custom generic object permission models

1.  Adds configuration settings `GUARDIAN_USER_OBJ_PERMS_MODEL` and `GUARDIAN_GROUP_OBJ_PERMS_MODEL`.  These should be set at the start of a project and not changed after.
```
GUARDIAN_USER_OBJ_PERMS_MODEL = 'articles.BigUserObjectPermission'
GUARDIAN_GROUP_OBJ_PERMS_MODEL = 'articles.BigGroupObjectPermission'
```

2. No changes for external libraries.  External libraries still import the same
```
from guardian.models import UserObjectPermission, GroupObjectPermission
```

3.  Internal to guardian the preferred method is the modified `get_user_obj_perms_model()` and `get_group_obj_perms_model()` which will return the generic object permission model when no DRF match or no object is passed.  Internal example:
```
from guardian.utils import get_user_obj_perms_model, get_group_obj_perms_model
UserObjectPermission = get_user_obj_perms_model()
GroupObjectPermission = get_group_obj_perms_model()
```

4. Custom user defined generic object permissions must inherit from `UserObjectPermissionAbstract` and `GroupObjectPermissionAbstract`.  This example shows redefining 'id' field and adding additional indexes.
```
from guardian.models import UserObjectPermissionAbstract, GroupObjectPermissionAbstract

class BigUserObjectPermission(UserObjectPermissionAbstract):
    id = models.BigAutoField(editable=False, unique=True, primary_key=True)
    class Meta(UserObjectPermissionAbstract.Meta):
        abstract = False
        indexes = [
            *UserObjectPermissionAbstract.Meta.indexes,
            models.Index(fields=['content_type', 'object_pk', 'user']),
        ]


class BigGroupObjectPermission(GroupObjectPermissionAbstract):
    id = models.BigAutoField(editable=False, unique=True, primary_key=True)
    class Meta(GroupObjectPermissionAbstract.Meta):
        abstract = False
        indexes = [
            *GroupObjectPermissionAbstract.Meta.indexes,
            models.Index(fields=['content_type', 'object_pk', 'group']),
        ]
```

5. No migrations for guardian models.  Changes in the user defined models, are contained in the user's application.

6. If `GUARDIAN_USER_OBJ_PERMS_MODEL` and `GUARDIAN_GROUP_OBJ_PERMS_MODEL` were changed after the start of a project, the old generic object permission tables would need to be manually copied to the new tables.  Below is just a potential example:
```
INSERT INTO articles_biguserobjectpermission (object_pk, content_type_id, user_id)
SELECT object_pk, content_type_id, user_id
FROM guardian_userobjectpermission;
INSERT INTO articles_biggroupobjectpermission (object_pk, content_type_id, group_id)
SELECT object_pk, content_type_id, group_id
FROM guardian_groupobjectpermission;
```
Original discussion that created this request #653


guardian/apps.py
```diff
@@ -7,5 +7,8 @@ class GuardianConfig(AppConfig):
     name = 'guardian'
 
     def ready(self):
+        # Must patch Group here since generic
+        # group permission model is definable
+        monkey_patch_group()
```
brianmay: I would think this should be conditional, same as the `monkey_patch_user` below... Yes, I know this wasn't the case before...

partimer: Making `monkey_patch_group()` call dependent on `GUARDIAN_MONKEY_PATCH = True` breaks the use case of `GuardianUserMixin` since `add_obj_perm()` and `del_obj_perm()` would not be defined on `django.contrib.auth.model.Group`.

To clarify `GuardianUserMixin` use case:
- custom user model
- custom user model is dependent on guardian library
- `GUARDIAN_MONKEY_PATCH = False` 
- `GuardianUserMixin` is mixed in with the custom user model
For example:
```
class CustomUser(GuardianUserMixin, AbstractUser):
    # Custom added fields  etc ....
    pass
```

_TL;DR My current understanding, not always calling monkey_patch_group() will break guardian._

brianmay: Wouldn't there be the same breakage not calling `monkey_patch_user()`? Or is that different?

example_project/articles/migrations/0002_custom_generic_permissions.py
```diff
@@ -0,0 +1,68 @@
+# Generated by Django 2.1.14 on 2019-11-14 10:41
```
auvipy: would be great if you could try this with django 2.2+

guardian/apps.py
partimer: No, when `monkey_user_patch()` is not called then custom user model must subclass `guardian.mixins.GuardianUserMixin`.

[Guardian Docs - Custom User Model](https://django-guardian.readthedocs.io/en/stable/userguide/custom-user-model.html#custom-user-model)

example_project/articles/migrations/0002_custom_generic_permissions.py
partimer: I would expect Django 2.1 support to break and TravisCI to fail.

Generating on 2.2 would make the example project migrations dependent `auth.0011_update_proxy_permissions` (Django >= 2.2) rather than `auth.0009_alter_user_last_name_max_length` (Django ~2.1).

brianmay: Also note the CI tests do  test Django 2.2.

partimer: @auvipy If it is ok with you, I would like to resolve this conversation.

partimer:
Is it possible to have this merged in?  That way it can be apart of @auvipy django 3.0 support changes.

guardian/utils.py
```diff
@@ -133,15 +133,33 @@ def get_40x_or_None(request, perms, obj=None, login_url=None,
...
+    """
+    try:
+        setting_value = getattr(guardian_settings, setting_name)
+        return django_apps.get_model(setting_value, require_ready=False)
+    except ValueError:
+        raise ImproperlyConfigured("{} must be of the form 'app_label.model_name'".format(setting_value))
+    except LookupError:
+        raise ImproperlyConfigured(
+            "{} refers to model '{}' that has not been installed".format(setting_name, setting_value)
+        )
```
michael-k: It would be nice to use [explicit exception chaining](https://www.python.org/dev/peps/pep-3134/#explicit-exception-chaining).

docs/configuration.rst
```diff
@@ -167,3 +167,66 @@ not be compatible with non-standard deployments, and should only be used when no
...
+      id = models.BigAutoField(editable=False, unique=True, primary_key=True)
+      class Meta(UserObjectPermissionAbstract.Meta):
+         abstract = False
+         indexes = [
+            *UserObjectPermissionAbstract.Meta.indexes,
+            models.Index(fields=['content_type', 'object_pk', 'user']),
+         ]
+
+
+Configure guardian to use the custom model in `settings.py`
```
michael-k:
```suggestion
Configure guardian to use the custom model in ``settings.py``
```

guardian/models/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+)
+
+# Must import after .models
+# When .models is loaded, default generic object permissions are created
+# The following statements may redirect external references to custom
+# generic object permission models
+from guardian.utils import get_user_obj_perms_model, get_group_obj_perms_model
+UserObjectPermission = get_user_obj_perms_model()
+GroupObjectPermission = get_group_obj_perms_model()
+
```
michael-k: Please add `__all__ = […]` to silence unused-import warnings.

docs/configuration.rst
```diff
@@ -167,3 +167,66 @@ not be compatible with non-standard deployments, and should only be used when no
 invocations would result in a large number of queries or when latency is particularly important.
 
 Defaults to ``False``.
+
+GUARDIAN_USER_OBJ_PERMS_USE
```
michael-k: Why not `GUARDIAN_USER_OBJ_PERMS_MODEL`?

partimer: Good idea.  _MODEL would be clearer than _USE.
Updated names in commit 15eb65d95fa50c23761d0343812b598553b1c0ee

guardian/utils.py
partimer: Added explicit exception chaining in 803a26cd87d3767be835947dc11f08406052a0c7

I can't think of any issues with the chaining the exception which is different than [django.contrib.auth.get_user_model()](https://github.com/django/django/blob/abc51d44afcb8482c64c0bb8c1022f11e7eb2bbb/django/contrib/auth/__init__.py#L151-L162)  (but that might just be left over from prior Python2 compatibility)

guardian/models/__init__.py
partimer: added in 72c71f05caae667de69a66925608e140fcbe34fc

Also, note Permission and Group are included only because some files within guardian were importing these from guardian.models rather than from django.contrib.auth.models.  They have been left in just in case some downstream library was doing the same.

example_project/articles/migrations/0002_custom_generic_permissions.py
partimer: Resolving conversation since requiring Django 2.2 would fail the current configuration of Travis CI

PR #661, opened 2019-12-09
staticfiles is depreceted since 2.1

l0rb:
`staticfiles` was deprecated in 2.1 an has been removed in 3.0

https://docs.djangoproject.com/en/3.0/releases/2.1/#deprecated-features-2-1

### DIFF
--- a/contrib/travis/test.sh
+++ b/contrib/travis/test.sh
@@ -15,5 +15,5 @@ if [ "${DJANGO_VERSION:0:3}" = "2.1" ] || \
     pip install .;
     cd example_project;
     python -Wa manage.py makemigrations --check --dry-run;
-    python -We manage.py test;
+    python -Wa manage.py test;
 fi;
--- a/contrib/travis/test.sh
+++ b/contrib/travis/test.sh
@@ -15,5 +15,5 @@ if [ "${DJANGO_VERSION:0:3}" = "2.1" ] || \
     pip install .;
     cd example_project;
     python -Wa manage.py makemigrations --check --dry-run;
-    python -Wa manage.py test;
+    python -We manage.py test;
 fi;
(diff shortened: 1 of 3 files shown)


ad-m:
Is there any warn to convert warnings to errors for our CI?

l0rb:
> Is there any way to convert warnings to errors for our CI?

I'm unsure how this relates to this issue? If django3 is added to the CI these kind of things should error in tests.

ad-m:
Since version 3.0, ```staticfiles``` will trigger an error without additional action. 

I am thinking about test-driven-development (failing test first, then fix it) approach to lead the current project to error first (we test on Django 2.1, so when deprecation warning convert to error somehow we should notice it as failing test) and then fix it.

I do not consider this a requirement for acceptance, which is why I accepted the change, but a nice, closely related element to prevent such problems in the future.

l0rb:
Happy to oblige. All warnings are now exceptions. Which of course makes the CI fail.

brianmay:
@ad-m It most definitely is possible to automatically turn warnings into exceptions. Have seen this on production, drives me crazy...

CI and tests is a really good place to use this however. Any warnings generated is probably something you want to try to avoid.

Not tested, but I think the two main ways are:

```
import warnings
warnings.filterwarnings("error")
```

And:

```
python -W error foobar.py
```

See https://stackoverflow.com/questions/5644836/in-python-how-does-one-catch-warnings-as-if-they-were-exceptions

brianmay:
The one problem with turning all warnings into errors for tests is that this will cause tests to fail for warnings caused by 3rd party packages. I have a suspicion this might be why tests are failing now (but I haven't verified this).

ad-m:
I have not considered that third-party libraries will generate problems for us (In NodeJS it usually just works). I think you have more experience in this area and I trust that you will make the right decision. If we could detect such problems in advance - it would be easier to maintain the library. However, if our ecosystem is not in good shape - we will have to abandon this idea.

l0rb:
I have reverted the commit that turned warnings into errors. It just makes all pipelines fail for reasons unrelated to the code in the PR. It's probably best to just try and always fix warnings.

ad-m:
@l0rb , merged. I created #663 to track proactive detection of deprecation warnings.

PR #675, opened 2020-02-05
Cast object_pk to UUIDField when PK is a UUID

richarddewit:
This fixes 

> ProgrammingError: operator does not exist: uuid = character varying

when having UUIDs as primary keys.

UUIDs used to work up until 2.1, but when I upgraded to 2.2 it threw the above error. PR #637 changed the way the `object_pk` was casted/compared, but only for Integer/AutoField PKs. This PR handles UUIDs as well.


tarikozket:
We are having the same problem.

ad-m:
@richarddewit , tests fail for MySQL. We can not accept PR which break tests for MySQL. Could you fix that?
@tarikozket , pull requests that eliminate the problem and work on all supported database engines with valid tests are welcome.

tarikozket:
> @tarikozket , pull requests that eliminate the problem and work on all supported database engines with valid tests are welcome.

@ad-m don't get me wrong, I just didn't know how to show that there is more people affected by this. Thank you for your great effort on the library!

ad-m:
Despite this, pull requests in the project are welcome, as well as comments about the code (review) of others :)

richarddewit:
@ad-m unfortunately MySQL doesn't support UUID fields (I wasn't aware) any good way I can check for this?

ad-m:
We can skip tests in this situation. As I remember we have some tests conditional already

ad-m:
Here is example of skipped tests: https://github.com/django-guardian/django-guardian/search?q=%22unittest.skipIf%22&unscoped_q=%22unittest.skipIf%22

richarddewit:
@ad-m thanks, I will try it on Monday

ad-m:
Please verify if MySQL supports UUID. The UUIDField field in Django (https://docs.djangoproject.com/en/3.0/ref/models/fields/#uuidfield) has no limit that it is only available in PostgreSQL, however the implementation differs from the database engine. 

I think the problem should be eliminated at source (if apply) rather than ignored by skipping the test.

richarddewit:
Ok I have no idea why it's going wrong on MySQL. When it's getting the UUID value back from the database, it's calling `convert_uuidfield_value` which passes back a 28-length string (like `b18d6e2c94184f90979f845ff39d`) to the UUID class which results in this error, because it expects a 32-length string (a UUID with the hyphens removed). I need to install MySQL to further investigate.

ad-m:
Can you explain if you need community support or are you going to solve the problem yourself in the near future? I make sure because in #688 users repeat status queries, and I don't see other parallel works to solve this issue.

richarddewit:
@ad-m - Yes sorry. I need the tests on MySQL to complete. We're using this fix for PostgreSQL in production, so it's stable. It's just the MySQL tests that are failing because it's doing something odd with UUID conversion.

guardian/shortcuts.py
```diff
@@ -617,28 +618,29 @@ def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=Fals
...
+    if cast_pk_field:
         values = values.annotate(
-            obj_pk=Cast(field_pk, BigIntegerField())
+            obj_pk=Cast(field_pk, cast_pk_field())
         )
         field_pk = 'obj_pk'
 
     values = values.values_list(field_pk, flat=True)
-    q = Q(pk__in=values)
+    q = Q(pk__in=list(values))
```
michael-k, line 633: Why are you evaluating the queryset here and in a few other places?

guardian/shortcuts.py
```diff
@@ -783,23 +785,36 @@ def get_objects_for_group(group, perms, klass=None, any_perm=False, accept_globa
...
+            BigIntegerField,
+            PositiveIntegerField,
+            PositiveSmallIntegerField,
+            SmallIntegerField,
+        ),
+    ):
+        return BigIntegerField
+
+    if isinstance(queryset.model._meta.pk, UUIDField):
+        return UUIDField
```
michael-k, line 818: Related import: `from django.db import connection`

```suggestion
    if isinstance(queryset.model._meta.pk, UUIDField) and connection.features.has_native_uuid_field:
        return UUIDField
```

michael-k:
Superseded by PR #691

PR #677, opened 2020-02-06
Fix error with get_objects_for_user on PostgreSQL.

### LINKED ISSUE #676: v2.2: Model Inheritance on PostgreSQL breaks get_objects_for_user
Hi there!

I've recently hit an exception using `guardian.shortcuts.get_objects_for_user` when upgrading from v2.1 to v2.2.

Our Django application uses a PostgreSQL backend and [model inheritance](https://docs.djangoproject.com/en/2.2/topics/db/models/#model-inheritance), similar to the following:

```py
# models.py
from django.db import models

class ParentTestModel(models.Model):
    created_on = models.DateTimeField(auto_now_add=True)

class ChildTestModel(ParentTestModel):
    parent_id = models.OneToOneField(
        ParentTestModel,
        on_delete=models.CASCADE,
        parent_link=True
    )
    name = models.CharField(max_length=31)
```

With this setup, the following code...

```py
get_objects_for_user(user, 'testapp.change_childtestmodel', ChildTestModel)
```

...causes an error. In our application, here's the exception:

```
Error
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 86, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UndefinedFunction: operator does not exist: integer = character varying
LINE 1: ...test-org' AND ("core_storeddocument"."element_id" IN (SELECT...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

Diving into the code, I have noticed [recent changes to get_objects_for_user](https://github.com/django-guardian/django-guardian/commit/7803d0c3afdf09232960b70310759ac131e29824#diff-00305e2706413da2ce53d91a9f9c3655R801-R805) by @woakas that includes the following function:

```py
def _is_cast_integer_pk(queryset):
    return isinstance(queryset.model._meta.pk, (
        IntegerField, AutoField, BigIntegerField,
        PositiveIntegerField, PositiveSmallIntegerField,
        SmallIntegerField))
```

I have a theory that this is the problem: because `ChildTestModel._meta.pk` will be of type `OneToOneField`, `_is_case_integer_pk` will be `False` and the conversions will not be put in place. If this is true, I think `_is_cast_integer_pk` will need extending to follow `ForeignKey` fields to determine if they are integer fields; something like:

```py
def _is_cast_integer_pk(queryset):
+   pk = queryset.model._meta.pk
+
+   if isinstance(pk, ForeignKey):
+       return _is_cast_integer_pk(pk.target_field)
+
    return isinstance(pk, (
        IntegerField, AutoField, BigIntegerField,
        PositiveIntegerField, PositiveSmallIntegerField,
        SmallIntegerField))
```

I [added a test](https://github.com/jonnyarnold/django-guardian/commit/6cda43b5ad77a7e7420d7e30d5117c6c19c5ad56) to verify that this is the issue, but the test passes 🤔 I believe this is because django-guardian's test application uses SQLite and SQLite doesn't mind doing the conversion. Unfortunately I don't know of a good way to switch DB backends in testing, or whether that's overkill for just one test.

---

In summary, I'd love it if:

- Someone could [pull my branch](https://github.com/jonnyarnold/django-guardian/tree/model-inheritance) and run the tests with a PostgreSQL backend to verify that the issue is what I have described.
- Someone could confirm that the problem is `_is_cast_integer_pk`, and refactor it to deal with `OneToOneField`s as primary keys.

Thanks!

jonnyarnold:
Fixes #676.

The first commit shows that there is indeed an issue affecting PostgreSQL. The second commit fixes it.


ad-m:
@woakas as contributor of changes of source of issue #676, could you provide feedback for that PR?

guardian/testapp/tests/test_shortcuts.py
```diff
@@ -1028,6 +1028,18 @@ def test_non_integer_primary_key_with_group_values(self):
...
+        assign_perm('testapp.change_childtestmodel', self.user, child_with_perm)
+        child_without_perm = ChildTestModel.objects.create(name="child2")
+
+        children = get_objects_for_user(self.user, 'testapp.change_childtestmodel', ChildTestModel)
+
+        self.assertEqual(1, len(children))
+        self.assertIn(child_with_perm, children)
+        self.assertNotIn(child_without_perm, children)
+
+
```
DavisRayM: Hey, relatively new to the community. Might be overstepping here(my bad if I am) seems like we have an extra line here ? Not sure if there is any linting standards that are forced onto the codebase... but this kinda looks weird when compared with the rest of the code....

_Once again.. my bad if I'm overstepping._

nikbora:
@woakas / @ad-m is there anything that can help move this along? I would like to upgrade to 2.2 but this prevents us from doing so in our project.

ad-m:
@nikbora , rebase please to make tests pass due ```pyupgrade``` auto-update, please. Immediately, when I see lights up green - I'm okay to merge the changes.

nikbora:
@jonnyarnold would be easier if you rebase on this PR branch rather than me fork again.

PR #687, opened 2020-03-08
Fix broken Django docs link

johnthagen:
Also update other links to point to latest stable docs rather than insecure versions of Django.


johnthagen:
@ad-m Anything else you'd like to see to get this merged?

ad-m:
No, I think I can merge it alone. We just don't have clear policy regarding accept number & merge.

PR #686, opened 2020-03-05
Modify 'assign_perm' shortcut to treat lists as a queryset

DavisRayM:
## Changes Implemented

- Modified the `bulk_assign_perm` function within the`BaseObjectPermissionManager` class to handle a list of objects
- Modified the `assign_perm` shortcut function to handle a list of objects
- Added @DavisRayM to the `AUTHORS` file

Fixes #610

PR #670, opened 2020-01-15
Release v2.2.0

michael-k:
@ad-m Can you create a new release?  2.1.0 is not compatible with Django 3.0 due to django-guardian/django-guardian#649 and django-guardian/django-guardian#661.

I tried to simplify that by getting the release notes started.

PR #691, opened 2020-04-14
Cast object_pk to UUIDField when PK is a UUID

michael-k:
This a rebase of PR #675 and also includes a fix for databases without a native uuid field (all but PostgreSQL).  The fix was inspired by https://github.com/django/django/commit/d9881a025c15d87b2a7883ee50771117450ea90d.

Supersedes PR #675


ad-m:
@michael-k , do you have a plan to create a new release? What do you need for that?

michael-k:
No concrete plans.  We could do a 2.2.1 release since all commits except PR #686 are fixes or documentation updates.

If we do a 2.3.0 release, I'd like to drop support for Django 2.1 (EOL since December) as announces in the release notes of 2.2.0.  Even 1.11 LTS has now reached its EOL (on April 1st) and there's no reason to support anything < 2.2 LTS.
https://www.djangoproject.com/download

lnagel:
Hey @michael-k - how could I help in getting a 2.2.1 release tagged?

The reason I'm asking is that I deployed an upgrade to Django 3.0 yesterday for a bigger project and we're stuck on 2.1.0 for the time being, but the admin views are not Django 3.0 compatible.

ad-m:
@Inagel, first step for any release is update changelog. See for example #670 . When everything in CHANGELOG I see no obstacles to create even small release. See also #694, which would allow us to further increase the frequency of releases.

lnagel:
I tried something, please see #695

lnagel:
@michael-k Thanks for this PR - confirmed fixed for my Django 3.0 app as well!

PR #693, opened 2020-04-14
Drop support for Django 2.1

michael-k:
Reached its EOL on 2019-12-02
https://www.djangoproject.com/download/#supported-versions

Should be included in 2.3.0, but not in 2.2.x. See https://django-guardian.readthedocs.io/en/stable/develop/changes.html#release-2-2-0-january-3-2020

PR #695, opened 2020-04-16
Update CHANGES for v2.3.0

lnagel:
I tried the following picking list for the release, so the cherry-pick commands would apply successfully:

```bash
git cherry-pick 9206ee4b32b3c97a7bc60440ce8221d483f1a24a
git cherry-pick 448a6149be8de47a62bd31c71eee1d94e688c5ef
git cherry-pick b12212babd8736015cd5b97a0a45dcc16c58c8a2
git cherry-pick 414d3a4a7bbd0ac4c2ddac85e0dcc098cf130ca1
git cherry-pick b338092f829b631cd59827bf29171f1ab405ff18
git cherry-pick 8e8dab207296ee37aa1d19eaeebfad7d0642f138
git cherry-pick 9c981804bbebe9d91f1d4d78292bedf292745986
git cherry-pick d816cfe4dad04dca78417d9c8613b49822256c60
```

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,4 @@
-Release 2.2.1 (January 16, 2020)
+Release 2.3.0 (January 16, 2020)
 ===============================
 
 * Fix formatting for pyupgrade
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,12 @@
 Release 2.2.1 (January 16, 2020)
 ===============================
 
+* Fix formatting for pyupgrade
+* Fix error with get_objects_for_user on PostgreSQL
+* Modify 'assign_perm' shortcut to treat lists as a queryset
+* Updated links to Django documentation and django-authority repository
+* Removed reference to "easy_install" in the documentation
+* Run migrations check only if required version of django is installed
 * Fix compatibility with UUIDField primary keys
 
 Release 2.2.0 (January 3, 2020)
--- a/CHANGES
+++ b/CHANGES
@@ -1,13 +1,14 @@
-Release 2.3.0 (January 16, 2020)
+Release 2.3.0 (May 6, 2020)
 ===============================
 
+* Drop support for Django 2.1
+* Fix compatibility with UUIDField primary keys
 * Fix formatting for pyupgrade
 * Fix error with get_objects_for_user on PostgreSQL
 * Modify 'assign_perm' shortcut to treat lists as a queryset
 * Updated links to Django documentation and django-authority repository
 * Removed reference to "easy_install" in the documentation
 * Run migrations check only if required version of django is installed
-* Fix compatibility with UUIDField primary keys
 
 Release 2.2.0 (January 3, 2020)
 ===============================
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,8 @@
+Release 2.2.1 (January 16, 2020)
+===============================
+
+* Fix compatibility with UUIDField primary keys
+
 Release 2.2.0 (January 3, 2020)
 ===============================
 


ad-m:
I kindly ask you to take into account all significant changes between the last release and the branch "devel" instead of just interesting for you. As a rule, the release is a snapshot of the "devel" branch. You can see also what pull requests was merged in meantime.

lnagel:
I'm sorry, I must have misunderstood. I assumed you were preparing for a 2.3.0 release in the devel branch and for that reason I did not include the django and documentation related changes.

ad-m:
For example PR #686 is also a change that has been merge in the "devel" branch since the last release, so it should be in the next django-guardian release, so it should be included in the CHANGELOG.

lnagel:
@ad-m Could you clarify whether your intention is to release everything in devel as 2.2.1 or as 2.3.0?

ad-m:
Definitely as to the next version symbol depends on the type of changes that have been made, which requires analyzing what changes have been made. Basically, we follow https://semver.org/, so we are open to suggestions after updating the changelog.

lnagel:
In this case there is new functionality: Modify 'assign_perm' shortcut to treat lists as a queryset

So this would mean that we go with 2.3.0?

ad-m:
It make sense. Regarding 2.3.0 @michael-k mention:

> If we do a 2.3.0 release, I'd like to drop support for Django 2.1 (EOL since December) as announces in the release notes of 2.2.0. 

In that situation release django-guardian 2.3.0 is blocked by #693.

lnagel:
I see. Then in this case would you either consider (a) releasing as 2.2.1; or (b) I could cherry-pick the commits to exclude "Modify 'assign_perm' shortcut to treat lists as a queryset" feature;

ad-m:
Why not focus on completing the changes in #693 and just release version 2.3.0? We do not have enormous pressure (for example, due to security issue) to disrupt the linearity of the software development process.

lnagel:
All clear. I renamed the version in this PR to 2.3.0. Let's discuss further in the #693 thread

lnagel:
If anyone else wants a release desperately, then please try to find time to follow @ad-m 's suggestion and review the PR #693 please.

I had to revert to using a custom built package with these commits cherry picked since I was not able to keep my customers waiting, they needed the fix to continue using the app.

CHANGES
```diff
@@ -1,3 +1,14 @@
+Release 2.3.0 (January 16, 2020)
+===============================
+
+* Fix formatting for pyupgrade
```
ad-m, line 4: Can we sort / group it in some way so that minor issues (formatting changes, minor documentation updates) do not cover the expected patches (fixing a bug in the UUID listed at the very end of the list)?

CHANGES
```diff
@@ -1,3 +1,14 @@
+Release 2.3.0 (January 16, 2020)
```
ad-m, line 1: Could you update date?

CHANGES
```diff
@@ -1,3 +1,14 @@
...
+===============================
+
+* Fix formatting for pyupgrade
+* Fix error with get_objects_for_user on PostgreSQL
+* Modify 'assign_perm' shortcut to treat lists as a queryset
+* Updated links to Django documentation and django-authority repository
+* Removed reference to "easy_install" in the documentation
+* Run migrations check only if required version of django is installed
+* Fix compatibility with UUIDField primary keys
+
```
ad-m, line 11: Can you update that PR to reflect latest changes?

CHANGES
lnagel: moved "Fix compatibility with UUIDField primary keys" up

CHANGES
lnagel: added "Drop support for Django 2.1"

lnagel:
@ad-m Please check also https://github.com/django-guardian/django-guardian/pull/700

lnagel:
@ad-m Are there any blockers to the release currently?

ad-m:
No, I haven't noticed your changes. Everything looks ready. In the evening I will release a new version.

PR #697, opened 2020-04-25
add any_perm parameter for PermissionRequiredMixin

1vank1n:


ad-m:
Why not adjust ```get_40x_or_None```? 

Missing tests.

1vank1n:
> Why not adjust get_40x_or_None?
Just have tried to "Keep It Stupid Simple".

> Missing tests.
Yeap. If it could be potential merged, I'll write them.

ad-m:
I understand that it was easy to make such a change without analyzing the whole thing, but it is more difficult to trace the code and maintain it in such a way, because similar logic is spread in two places.

1vank1n:
Ok! I'll rewrite to adjust `get_40x_or_None`

1vank1n:
@ad-m what about this build?

guardian/utils.py
```diff
@@ -106,8 +107,12 @@ def get_40x_or_None(request, perms, obj=None, login_url=None,
...
     # if still no permission granted, try obj perms
     if not has_permissions:
-        has_permissions = all(request.user.has_perm(perm, obj)
-                              for perm in perms)
+        if any_perm:
+            has_permissions = any(request.user.has_perm(perm, obj)
+                                for perm in perms)
+        else:
+            has_permissions = all(request.user.has_perm(perm, obj)
+                                for perm in perms)
```
michael-k, line 115:
```suggestion
            has_permissions = any(request.user.has_perm(perm, obj)
                                  for perm in perms)
        else:
            has_permissions = all(request.user.has_perm(perm, obj)
                                  for perm in perms)
```

PR #706, opened 2020-06-17
Use chain.from_iterable in core.py

cool-RR:
This is a faster and more idiomatic way of using `itertools.chain`. Instead of computing all the items in the iterable and storing them in memory, they are computed one-by-one and never stored as a huge list. This can save on both runtime and memory space.

### DIFF
--- a/guardian/core.py
+++ b/guardian/core.py
@@ -158,9 +158,7 @@ class ObjectPermissionChecker:
                 return []
             if self.user and self.user.is_superuser:
                 perms = list(
-                    chain.from_iterable(
-                        Permission.objects.filter(content_type=ctype).values_list("codename")
-                    )
+                    Permission.objects.filter(content_type=ctype).values_list("codename", flat=True)
                 )
             elif self.user:
                 # Query user and group permissions separately and then combine
@@ -195,9 +193,7 @@ class ObjectPermissionChecker:
 
         if self.user and self.user.is_superuser:
             perms = list(
-                chain.from_iterable(
-                    Permission.objects.filter(content_type=ctype).values_list("codename")
-                )
+                Permission.objects.filter(content_type=ctype).values_list("codename", flat=True)
             )
 
             for pk in pks:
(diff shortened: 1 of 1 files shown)


cool-RR:
@inagel What do you think?

guardian/core.py
```diff
@@ -157,9 +157,11 @@ def get_perms(self, obj):
...
                 return []
             if self.user and self.user.is_superuser:
-                perms = list(chain(*Permission.objects
-                                   .filter(content_type=ctype)
-                                   .values_list("codename")))
+                perms = list(
+                    chain.from_iterable(
+                        Permission.objects.filter(content_type=ctype).values_list("codename")
+                    )
+                )
```
michael-k, line 164: The simpler solution is to pass `flat=True` to `values_list`.
```suggestion
                perms = list(
                    Permission.objects.filter(content_type=ctype).values_list("codename", flat=True)
                )
```

guardian/core.py
```diff
@@ -192,10 +194,11 @@ def prefetch_perms(self, objects):
...
         if self.user and self.user.is_superuser:
-            perms = list(chain(
-                *Permission.objects
-                .filter(content_type=ctype)
-                .values_list("codename")))
+            perms = list(
+                chain.from_iterable(
+                    Permission.objects.filter(content_type=ctype).values_list("codename")
+                )
+            )
```
michael-k, line 201:
```suggestion
            perms = list(
                Permission.objects.filter(content_type=ctype).values_list("codename", flat=True)
            )
```

guardian/core.py
```diff
@@ -245,9 +248,7 @@ def prefetch_perms(self, objects):
             group_perms_qs = group_model.objects.filter(**group_filters).select_related('permission')
             perms = chain(user_perms_qs, group_perms_qs)
         else:
-            perms = chain(
-                *(group_model.objects.filter(**group_filters).select_related('permission'),)
-            )
+            perms = group_model.objects.filter(**group_filters).select_related('permission')
```
michael-k, line 251: Nice catch!

cool-RR:
> Are those lists ever huge? Since the generator returned by `chain`/`chain.from_iterable` is directly passed to `list()`, I don't think this could save any memory.

That's mostly true, it'll save memory only in the short interval when the list is being built and the `*args` tuple hasn't been deallocated yet. It's more of a general best practice thing than an optimization in this case. I have no reason to think these lists are huge.

PR #715, opened 2020-09-27
Fix function reference in docs: userguide remove section

pedrorojasg:
Greetings, I propose to update the "assign" reference to "assign_perm".
To be consistent with the new version of the function and the rest of the docs that directly reference assign_perm.


michael-k (approved):
LGTM

PR #716, opened 2020-09-30
Improved README.rst

rims2020:


README.rst
```diff
@@ -61,6 +61,9 @@ We need to hook ``django-guardian`` into our project.
 
      python manage.py migrate
 
+4. Create superuser for viewing the admin page.
+     python manage.py createsuperuser
```
michael-k, line 65: This is not specific to guardian.  I'd prefer a general note about the admin with a link to https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#overview instead.

michael-k (changes requested):

README.rst
RainshawGao: I agree with @michael-k .

thclark:
I agree this is basic django knowledge; closing the PR.

PR #717, opened 2020-10-08
Fixed minimum Django version number.

roskakori:
There seems to be typo in the version number, Django 1.2 would be ancient.

### DIFF
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -31,7 +31,7 @@ Alternate projects
 
 Django_ still *only* has the foundation for object permissions [1]_ and
 ``django-guardian`` makes use of new facilities and it is based on them.  There
-are some other pluggable applications which do *NOT* require Django_ version 1.2+. 
+are some other pluggable applications which do *NOT* require Django_ version 2.2+. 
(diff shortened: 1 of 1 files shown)


michael-k:
It's not a typo, it's just very old and hasn't been touched in 3 years.  I think the whole paragraph should be adjusted.

roskakori:
> It's not a typo, it's just very old and hasn't been touched in 3 years. I think the whole paragraph should be adjusted.

Hm, then it's a bit confusing in combination with the following sentence in `installation.rst`:

> This application requires Django 2.2 or higher

So what the docs try to say is that django-guardian requires Django 2.2+, while some other libraries require a Django version before 1.2, which might be a major selling point to some people?

ad-m:
This sentence has not been updated in years. It requires a general review to see if it still provides relevant information. I think the reference to 1.2+ is no longer relevant to anybody. We can remove this sentence. Perhaps, however, it is worth considering changing the entire paragraph, because alternative libraries have also appeared.

michael-k:
I propose to close this due to inactivity.

PR #721, opened 2020-10-15
Fixed default_app_config deprecation

smithdc1:
Django 3.2 automatically detects `AppConfig` and therefore this setting is no longer required.  Here is the warning in the CI and the release notes. 

https://docs.djangoproject.com/en/dev/releases/3.2/#automatic-appconfig-discovery

https://travis-ci.org/github/django-guardian/django-guardian/jobs/735548344#L427

### DIFF
--- a/guardian/__init__.py
+++ b/guardian/__init__.py
@@ -2,8 +2,10 @@
 Implementation of per object permissions for Django.
 """
 from . import checks
+import django
 
-default_app_config = 'guardian.apps.GuardianConfig'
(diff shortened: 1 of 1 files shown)


michael-k (approved):
LGTM

adamchainz:
Can we see a release? This warning is now breaking the django-rest-framework CI since elevate warnings to errors there (experimentally, to see if we can avoid passing warnings on to users).

PR #724, opened 2020-10-28
Make @permission_required work on methods

### LINKED ISSUE #723: @permission_required does not work on methods
The `@permission_required` decorator only works on view functions, not methods.
Specifically, [django-rest-framework](https://www.django-rest-framework.org/) lets you define ViewSets, which are classes that make it easy to quickly define a family of views for a model. Instead of defining individual view functions, you create a class which extends ViewSet, and create view methods on that class.

I would like to be able to do something like this:
```
class MyViewSet(ReadOnlyModelViewSet):
    ...
    @action(methods=['POST'])
    @permission_required('top_secret_clearance')
    def customize(self, request, pk):
        ... do the thing ...
```

However, the `permission_required` decorator is defined like this:
```
def permission_required(perm, lookup_variables=None, **kwargs):
    def decorator(view_func):
        def _wrapped_view(request, *args, **kwargs):
            ... permission checking ...
```
`_wrapped_view` only takes one argument, `request`, but when you use `@permission_required` on a method (not a function), it is passed `self, request` instead. `_wrapped_view` then assumes that `self` is actually `request`, and `request` is just one of the `*args`, and then fails when the ViewSet is not a request object.

dchiquito:
Fixes #723 

Make @permission_required work on methods
    
`@permission_required` was designed for methods that look like:
```
@permission_required('top_secret_clearance')
def my_view_function(request, arg1, arg2):
    ...
```
However, it would fail when used on methods, as the method signature
must have `self` as the first argument, like this:
```
class MyClass:
    @permission_required('top_secret_clearance')
    def my_view_method(self, request, arg1, arg2):
        ...
```

`@permission_required` will now accept a method signature like
`(request, *args, **kwargs)` or like `(self, request, *args, **kwargs)`.
If neither of the first two `*args` to the wrapper are an instance of
`HttpRequest`, an informative error is thrown.

### DIFF
--- a/guardian/decorators.py
+++ b/guardian/decorators.py
@@ -4,6 +4,7 @@ from django.contrib.auth import REDIRECT_FIELD_NAME
 from django.db.models import Model
 from django.db.models.base import ModelBase
 from django.db.models.query import QuerySet
+from django.http.request import HttpRequest
 from django.shortcuts import get_object_or_404
 from django.utils.functional import wraps
 from guardian.exceptions import GuardianError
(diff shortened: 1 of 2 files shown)


RainshawGao:
have you try to use `django.utils.decorators.method_decorator` ?
like in https://docs.djangoproject.com/zh-hans/3.1/topics/class-based-views/intro/#id1, use method_decorator with permission_required may work

RainshawGao:
As #723 is solved and has been closed by author, I think this PR should also be closed. 😊

PR #728, opened 2020-12-18
Respect GUARDIAN_GET_CONTENT_TYPE during ctype match check in guardian.shortcuts.get_objects_for_user

### LINKED ISSUE #727: Check in get_objects_for_user does not respect GUARDIAN_GET_CONTENT_TYPE
There is an equality check in https://github.com/django-guardian/django-guardian/blob/devel/guardian/shortcuts.py#L533 which fails when `GUARDIAN_GET_CONTENT_TYPE` is set.

The check in question is:
```
if ctype.model_class() != queryset.model
```

This works in a standard case, however if `GUARDIAN_GET_CONTENT_TYPE` is used and the expected content type for the model is not the content type for the model itself, this will fail.  My suggestion is to replace this check with:
```
if ctype != get_content_type(queryset.model):
```

I'm happy to create a PR for this, but wanted to pose it for feedback first.

markedwards:
- Closes #727

### DIFF
--- a/guardian/shortcuts.py
+++ b/guardian/shortcuts.py
@@ -530,7 +530,7 @@ def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=Fals
         raise WrongAppError("Cannot determine content type")
     else:
         queryset = _get_queryset(klass)
-        if ctype.model_class() != queryset.model:
+        if ctype != get_content_type(queryset.model):
             raise MixedContentTypeError("Content type for given perms and "
                                         "klass differs")
 


ad-m:
Could you add tests to avoid regression?

markedwards:
I realised there is the same issue in `get_objects_for_group`, so I made the same change there.  Also, added tests to confirm that calls to `get_objects_for_user` and `get_objects_for_group` do not raise when `GET_CONTENT_TYPE` is overridden.

ad-m (approved):

RainshawGao (approved):

PR #731, opened 2021-03-01
feat: added message for PermissionRequiredMixin

Rainshaw:
As in `django.contrib.auth.mixins.PermissionRequiredMixin`, there is a message passed to the PermissionDenied exception when permissions check fails.

This PR adds `object_permission_denied_message` to `guardian.mixins.PermissionRequiredMixin` and uses it in `get_object_permission_denied_message()` function. 

There is already a PR #720 which trys to add `permission_denied_message`, the reason I use `object_permission_denied_message` is that in this way we can use `django.contrib.auth.mixins.PermissionRequiredMixin` and `guardian.mixins.PermissionRequiredMixin`  at same time, such as

```
from django.contrib.auth.mixins import PermissionRequiredMixin
from guardian.mixins import PermissionRequiredMixin as ObjectPermissionRequiredMixin
class UpdateView(PermissionRequiredMixin, ObjectPermissionRequiredMixin, UpdateView):
    model = Client
    form_class = ClientForm
    template_name = 'client_update.html'
    permission_required = 'clients.change_client'
    permission_denied_message = 'You do not have the permission to change client'
    object_permission_denied_message = 'You are not the creator of this client, '
                                                                  'so you do not have the right to modify this client!'
    raise_exception = True

```
in this situation, if user do not have `clients.change_client`, he will see `permission_denied_message`.
else if user have `clients.change_client`, then will check `change_client` on the object, if false, he will 
see `object_permission_denied_message`.


thclark (approved):
I like this but I'd like to see a test that PermissionDenied('') is the same as PermissionDenied(), because if not then this is a breaking change (some people may rely on default error message)

PR #732, opened 2021-03-01
feat: register guardian settings globally

### LINKED ISSUE #726: Should settings.ANONYMOUS_USER_NAME be explicitly defined ?
I had the impression that if not defined settings.ANONYMOUS_USER_NAME would be defined by default by guardian and we can use this setting anywhere. 

At present I'm having an issue where this variable is not explicitly defined in my settings file but the code trying to use it is failing. 

Looking at `conf/settings.py` I see the default value being set but it is assigned to a variable with scope only in that module, not assigned as an attribute to the global settings module.

Similar to https://github.com/django-guardian/django-guardian/issues/480

IMO this is an error (or at lest misaligned documentation) and could be fixed with a simple `settings.ANONYMOUS_USER_NAME = ` assignment if not set. Please advise.

RainshawGao:
fix  #726

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -62,3 +62,4 @@ Authors ordered by first contribution
 - Davis Raymond Muro <davisraymondmuro@gmail.com>
 - Richard de Wit <henk.exe@gmail.com>
 - Pedro Rojas Gavidia <pedrorojas.gavidia@gmail.com> @pedrorojasg
+- Rainshaw Gao <rxg@live.com>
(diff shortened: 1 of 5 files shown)


dmgolembiowski:
Thank you; the team at my day-job has a vested interest in this PR.

ad-m:
Which other Django project perform something like that? In my project usually - if needed other project settings - I import that setting from that project.

RainshawGao:
@ad-m actually none but IMO this is really helpful, and issue #726 need this.

ad-m:
I believe there is now a better solution (see my comment in #726) than modifying the global object, which can have unpredictable and difficult to debug consequences. 

However, I am open to discussions. This is a popular need in Django libraries, so I'd like us to follow the ecosystem rather than come up with a new approach. So the easiest way to convince me is with examples of a solution that is widely adopted.

PR #749, opened 2021-05-26
Update tested Django versions to latest patch releases

### LINKED ISSUE #746: Test against and update classifiers for final Django 3.2 release
I noticed that the tox configuration already includes Django 3.2 RC1. It would be lovely if it is updated to test against 3.2 final, as well as if the tests pass, that the classifiers are updated to include Django 3.2 (final).

johnthagen:
Closes #746

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
-        django: [2.2.23, 3.0.14, 3.1.11, 3.2.3]
+        django: [2.2.17, 3.0.11, 3.1.3, 3.2.0]
         include:
           - python-version: 3.8
             django: main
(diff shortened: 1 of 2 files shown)


johnthagen:
@ad-m Would you be able to review this? I ensures that we are testing on a final release of Django 3.2.

michael-k:
This is a bit more than the what #746 requested.  The lower bounds for Django 2.2 to 3.1 are only there to ensure compatibility with Python 3.9.  `django~=2.2.17` etc. already installs the latest 2.2 release (at least on GitHub Actions because we don't use a cache there).

I'm :+1: for a bump of `django~=3.2rc1` to `django~=3.2.0`.  But I don't see the necessity to constantly bump the minimal patch release.

johnthagen:
Good point, I should have noticed the `~=`. I've reverted all but the 3.2.0 bump.

If this is accepted, please squash the commit while merging.

PR #754, opened 2021-07-06
Add type annotations

MartinThoma:


MartinThoma:
You can let mypy check the code with `mypy .`. It's very inefficient right now as it will only check functions where something in the signature has a type annotation. You can see way more with `mypy . --strict`.

thclark:
I actually would really welcome someone picking this up and taking it to the finish line!

dpgraham4401:
oooh, I'd really like to pick this one up (maybe before I finish #849). I have a better understanding of the types after having spending all that time with the docstring in #828.

thclark:
Closing in favour of #851

Issue #756, opened 2021-08-12
GuardedModelAdminMixin and GuardedModelAdmin are unsafe


areyal:
Any user who gets into any of the paths provided by GuardedModelAdminMixin can view, add, change and delete guardian permissions for any user, even if the current user has not guardian permissions.
```
PATH_TO_OBJECT/OBJECT_ID/permissions/
PATH_TO_OBJECT/OBJECT_ID/permissions/user-manage/USER_ID/ 
PATH_TO_OBJECT/OBJECT_ID/permissions/group-manage/GROUP_ID/ 
```
I think the problem lies in the methods which provide the views: they are not checking guardian permissions:
```
obj_perms_manage_view
obj_perms_manage_user_view
obj_perms_manage_group_view
```
I will try to make a fix

areyal:
This change works for me, but it is a bit rough.
Just check permissions `..or (not request.user.has_perm('guardian.view_userobjectpermission') and not request.user.has_perm('guardian.view_groupobjectpermission'))`

A complete solution would provide a modified template and show only users if has view_userobjectpermission permission but not view_groupobjectpermission.


```
    def obj_perms_manage_view(self, request, object_pk):
        """
        Main object permissions view. Presents all users and groups with any
        object permissions for the current model *instance*. Users or groups
        without object permissions for related *instance* would **not** be
        shown. In order to add or manage user or group one should use links or
        forms presented within the page.
        """
        if not self.has_change_permission(request, None) or (not request.user.has_perm('guardian.view_userobjectpermission') and not request.user.has_perm('guardian.view_groupobjectpermission')):
            post_url = reverse('admin:index', current_app=self.admin_site.name)
            return redirect(post_url)

        from django.contrib.admin.utils import unquote
        obj = get_object_or_404(self.get_queryset(
            request), pk=unquote(object_pk))
        users_perms = OrderedDict(
            sorted(
                get_users_with_perms(obj, attach_perms=True,
                                     with_group_users=False).items(),
                key=lambda user: getattr(
                    user[0], get_user_model().USERNAME_FIELD)
            )
        )

        groups_perms = OrderedDict(
            sorted(
                get_groups_with_perms(obj, attach_perms=True).items(),
                key=lambda group: group[0].name
            )
        )

        if request.method == 'POST' and 'submit_manage_user' in request.POST:
            user_form = self.get_obj_perms_user_select_form(
                request)(request.POST)
            group_form = self.get_obj_perms_group_select_form(
[...]

PR #757, opened 2021-08-13
Security bugfix: GuardedModelAdminMixin did not check for guardian pe…

areyal:
…rmissions. Issue #756


johnthagen:
Should this PR also have updated the documentation?

areyal:
> Should this PR also have updated the documentation?

Don't think so since it's just a fix

PR #763, opened 2021-10-23
Add support for Python 3.10

johnthagen:
Django 3.2 is currently the only Django release to support Python 3.10: https://docs.djangoproject.com/en/3.2/faq/install/#what-python-version-can-i-use-with-django

Django 3.2.9 is needed to fully support Python 3.10 (and fixes `distutils` `DeprecationWarning`). This PR should be ready around November 1 when 3.2.9 is released: https://docs.djangoproject.com/en/4.0/releases/3.2.9/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,11 +19,11 @@ jobs:
         include:
           - python-version: "3.10"
             django: 3.2.0
-          - python-version: 3.8
+          - python-version: "3.8"
             django: main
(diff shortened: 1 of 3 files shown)


johnthagen:
@ad-m This PR is ready now. Django 3.2.9 has been released which fixed the 3.10 issue that was causing CI to fail before.

ad-m:
@johnthagen thank you for keeping that project healthy.

Issue #766, opened 2021-11-25
Assign permission to any object of a type


NaturalBornCamper:
I'm trying to make an "admin" group with django-guardian that has access to all objects of a type instead of a single object, like such:

```python
from guardian.models import UserObjectPermission
from django.contrib.contenttypes.models import ContentType
from apps.myapp.models import Website

# Create groups that can edit websites (admin can edit every website, the other group only a specific website)
admins = Group.objects.create(name='admins')
website_44_editors = Group.objects.create(name='website_44_editors')

# This works, assigning permission to edit website 44 to the correct group
website_44 = Website.objects.get(pk=44)
UserObjectPermission.objects.assign_perm('change_website', website_44_editors, obj=website_44)
website_44_editors.has_perm('change_website', website_44)

####### This doesn't work, as an object is needed #######
UserObjectPermission.objects.assign_perm('change_website', admins)
# Thus, this would also work, without admins having permission on specific website
admins.has_perm('change_website', website_44)
```

Any way to do this? Or I should not be using Guardian for that second part?

ad-m:
What do you want is not object-level permission, but model-level permission. Model-level permission is build-in in Django, so you do not need Django-guardian for that. If you really want to use django-guardian you might use it like that:

```python
from guardian.shortcuts import assign_perm
assign_perm('articles.view_article', user, obj=None)
```

If you think that the issue has been resolved - please close it.

NaturalBornCamper:
I think I really need object-level permission. Because

- a `Website` with id 44 would be available to all users in the group `website_44_editors`
- a `Website` with id 56 would be available to all users in the group `website_56_editors`
- All `Website` object can be changed by the group `admins`

Furthermore, to see if a user has access to edit a certain article for example, we check if that user has permission to edit the website this article is attached to.

All we need is to check if a user has access to that `Website`, then if so, it can do anything to any model related to this `Website`

ociule:
Regarding this requirement, ad-m means it will work using assign_perm or you can do it through the admin site by adding the `articles.view_website` perm to the admin group:

```py
####### This doesn't work, as an object is needed #######
UserObjectPermission.objects.assign_perm('change_website', admins)

# This will work using assign_perm
from guardian.shortcuts import assign_perm
assign_perm('articles.view_website', admins, obj=None)
```
I think you need the app name as a prefix in the permission, btw.

[...]

PR #767, opened 2021-11-28
Use shortcuts.assign_perm in example

ad-m:
We have a nice helper `guardian.shortcuts.assign_perm`. It abstracts module-level permission (see #766) and direct foreign keys. I personally use it everywhere, so I suggest update example to encourage users to do the same.


iamMHZ (approved):

thclark (approved):
Seems like a no-brainer, thank you ad-m. @cmaggiulli Another easy decision here!

PR #768, opened 2021-12-11
union users and groups for better perf

qeternity:
In our benchmarks, the current `get_users_with_perms` shortcut is improved by unioning when using the `with_group_users` flag. For our typical usecases, this drops query time by ca. 60% (5ms -> 2ms) with reasonably large production tables running on postgres, and cut the number of queries in half (as we do not force execution of the group id query, which I suspect was done for performance reasons anyway).

Of course, I am sure that other database query planners may perform differently, but it is likely that the union approach still comes out ahead sufficiently by avoiding an outer join on potentially large tables.

### DIFF
--- a/guardian/shortcuts.py
+++ b/guardian/shortcuts.py
@@ -295,12 +295,15 @@ def get_users_with_perms(obj, attach_perms=False, with_superusers=False,
             }
         else:
             user_filters = {'%s__content_object' % related_name: obj}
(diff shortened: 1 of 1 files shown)


ad-m (approved):
It looks good to me, but I haven't run any benchmarks.

thclark (approved):
This makes sense to me, avoiding that set() call is probably the performance enhancer here.

PR #769, opened 2022-01-20
Update shortcuts.py

idiomaticrefactoring:
I refactored code with set comprehension which is more concise and efficient，how do think this change which has practical value？
refactoring code with set comprehension which is more concise and efficient

### DIFF
--- a/guardian/shortcuts.py
+++ b/guardian/shortcuts.py
@@ -552,9 +552,7 @@ def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=Fals
     has_global_perms = False
     # a superuser has by default assigned global perms for any
     if accept_global_perms and with_superuser:
-        for code in codenames:
-            if user.has_perm(ctype.app_label + '.' + code):
-                global_perms.add(code)
+        global_perms = {code for code in codenames if user.has_perm(ctype.app_label + '.' + code)}
         for code in global_perms:
             codenames.remove(code)
         # prerequisite: there must be elements in global_perms otherwise just follow the procedure for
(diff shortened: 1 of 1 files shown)


ad-m:
Could you remove redundant line `global_perms = set()` too?

idiomaticrefactoring:
Okay, I remove the `global_perms = set()`.

guardian/shortcuts.py
```diff
@@ -548,13 +548,10 @@ def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=Fals
...
-    global_perms = set()
     has_global_perms = False
     # a superuser has by default assigned global perms for any
     if accept_global_perms and with_superuser:
-        for code in codenames:
-            if user.has_perm(ctype.app_label + '.' + code):
-                global_perms.add(code)
+        global_perms = {code for code in codenames if user.has_perm(ctype.app_label + '.' + code)}
         for code in global_perms:
             codenames.remove(code)
```
sevdog, line 556: It could also be pointed that this operation may be rewritted using [set operators](https://docs.python.org/3/library/stdtypes.html#frozenset.difference):

```python
codenames = codenames - global_perms
```

thclark (approved):
This looks cleaner to me, it'll also be slightly faster than the original code. @cmaggiulli another no-brainer!
@idiomaticrefactoring and @sevdog thanks for your contributions on this - if you'd like to make that suggested tweak to do lines 555-556 as a set operation that'll be great, otherwise we'll proceed and merge anyhow.

PR #771, opened 2022-02-04
Derive auth backend from base django auth banckend

sevdog:
Default methods for authentication backends are already defined in [Basebackend](https://github.com/django/django/blob/11661d3815f273a6968cf09ecc678cf6fa6dcfcd/django/contrib/auth/backends.py#L8).

It is better to derive from this class than to add dummy methods to cover them.

### DIFF
--- a/guardian/backends.py
+++ b/guardian/backends.py
@@ -1,4 +1,5 @@
 from django.contrib.auth import get_user_model
+from django.contrib.auth.backends import BaseBackend
 from django.db import models
 from guardian.conf import settings
 from guardian.core import ObjectPermissionChecker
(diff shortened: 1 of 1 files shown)


iamMHZ (commented):
@sevdog 

`django.contrib.auth.backends.BaseBackend`   has    `get_user_permissions` and
`get_group_permissions` methods. Aren't they called in any other places? 
shouldn't they be implemented (maybe simply by calling guardian shortcuts)?

sevdog:
I found that this may cause problems with test-client `.force-login` method, which is going to use the first backend with a `.get_user` method: https://github.com/django/django/pull/7634


So this should not be merged.

PR #777, opened 2022-05-09
Allow batch assignment of multiple permissions to multiple users/groups for multiple objects

JCourt1:
Currently, the methods in the `BaseObjectPermissionManager` do not allow assigning multiple permissions for multiple users and for multiple objects. Likewise, the `assign_perm` function in shortcuts does not even allow assigning multiple permissions at once. The problem with this is that permission assignments are very slow, because they cannot be properly batched together.

This PR adds in methods to the object permission manager, and also adds two new shortcuts: `bulk_assign_perms` and `bulk_remove_perms`. In addition, I have also added a `commit` parameter, which allows the user of the shortcuts to go even further with the batching, and accumulate permissions so that they can all be persisted in one fell swoop. I have ensured that previous behaviour is left intact, and the new behaviour is well tested. 

Having applied these changes in a work project which relies heavily on django guardian, I have seen a big decrease in the time taken for the entire test suite to complete (down to ~20 seconds from 3 minutes).


Laityned:
I like this future, could this be merged?

CMarcher-Ahuora:
Is there any work that is going to be done to get this feature integrated with the current state of the repo? I'd love to use Guardian, but the lack of permission batching makes it completely unusable for my case.

thclark:
@CMarcher-Ahuora this work isn't planned by the core team, but if you care to raise an issue that discusses your use case, and are willing to resurrect this PR to make it happen, we'd happily consider adding it to the v3.1 milestone.

thclark:
Update: some work toward this has been undertaken in 3.3.0 release (imminent).

Contributions here are welcome, still outstanding documentation and ideally some benchmarks.

Natgho:
https://github.com/django-guardian/django-guardian/issues/1026

PR #790, opened 2022-10-02
Upating save() function for better performance

Muditalbet:
Updating save function of Django to tell it what s updated and not to update the complete model for particular field.


iamMHZ (approved):

thclark (approved):

thclark:
@cmaggiulli if you wouldn't mind reviewing this it seems like a no brainer :)

PR #793, opened 2023-02-08
add dropdown menu for users and groups before editing object permission in…

duwangthefirst:
- update only one single file: guardian/admin.py
- minimal code update
- achieved goal: allow you select user or group with dropdown menu (instead of searching User with username and searching Group by group's name) before edit user-object-permission or group-object-permission.


thclark:
Really appreciate the contribution @duwangthefirst but as @sevdog says this can become a performance problem, so I'm going to close this one. Thank you anyhow!

PR #803, opened 2023-09-01
Update managers.py to include get_prep_value()

gabn88:
Rebased from the devel branch.

This takes into account that if Model Fields use get_prep_value, that that value is used.


thclark:
Whilst I like the idea, unfortunately the implementation here requires use of `'id'` field as the `pk` which whilst common isn't  always the case. The use of `object.pk` is designed to pull the primary key field irrespective of its name. So, I'll close this for now but thanks anyhow for your contribution @gabn88

PR #814, opened 2024-03-17
ignore_conflicts on bulk create

pgiki:
ignore_conflicts when bulk creating to assign permissions. In most (all) cases, I have found these conflicts would not need special attention so I think it will be ok to simply ignore them


thclark:
This would be a breaking change @pgiki so I'd be very reluctant to do this without a clear use case - could you clarify your use case a bit better?

Alternatively perhaps the patch could be refactored to maintain the same `ignore_conflicts=False` default behaviour, but allow you to override that parameter from the caller.

PR #800, opened 2023-08-07
Add support for official Python/Django versions

irtazaakram:
Hi, 

This PR

* Add support for Django 4.1 & 4.2
* Add support for Python 3.10 & 3.11
* Drop support for Django 2.2, 3.0, 3.1 & 4.0
* Drop support for Python 3.5, 3.6 & 3.7

Please can we have a new release after this PR is merged.

Thanks,

PR #812, opened 2024-02-03
"Fix" performance enhancement related to #811

cmaggiulli:
Please see [Git Issue 811](https://github.com/django-guardian/django-guardian/issues/811) for a description of the problem

Issue #603, opened 2019-02-03
Django-guardian welcomes your support


ad-m:
In recent months, the project did not have new releases. This was due to the lack of time on the part of the current project maintainer. Even I myself do not have time, and I do not even use django-guardian in any actively developed project. 

This situation is not good for the project. It is necessary to devote some time for:

- reviewing outstanding issues to resolve them,
- helping people who make pull requests to the project to bring these changes to be consistent with the whole project,
- answering questions on Stackoverflow, because they appear there.
- dropping old-versions
- etc.

Who has a certain amount of time to carry out these tasks and lead the project further?

kiran-capoor94:
Hi, I’m new here. Let me know how I can help.

johnthagen:
Might be worth [pinning this issue](https://help.github.com/en/github/managing-your-work-on-github/pinning-an-issue-to-your-repository) for more visibility.

brianmay:
@johnthagen I didn't know you could do that. Thanks. Now done.

@kiran-capoor94 I noted that nobody replied to your request, at least not here. If you are still interested in helping, please see the list in the first message.

In particular, there are 16 pull requests open at present. Help reviewing these and fixing them up to a acceptable standard appreciated.

DavisRayM:
Hey 👋 ,  I would like to help around where I can too. Picking up a few issues and contributing code for now.

kiran-capoor94:
> @johnthagen I didn't know you could do that. Thanks. Now done.
> 
> @kiran-capoor94 I noted that nobody replied to your request, at least not here. If you are still interested in helping, please see the list in the first message.
> 
> In particular, there are 16 pull requests open at present. Help reviewing these and fixing them up to a acceptable standard appreciated.

Thanks, will review.

megaleunam:
I would like to help with some problems. i am currently using django-guardian on a project

ad-m:
@melvyn-sopacua , there are 13 pull requests open at present. Help reviewing these and fixing them up to a acceptable standard appreciated. Notes that you don't see problems are also support. There are also open development issues, e.g. #694. Everybody can take it and provide pull requests to solve it.

RainshawGao:
maybe you can consider jazzband.co

thclark:
@ad-m would you like to add me as a maintainer? Maybe give me pypi access too... I could at least triage some of the issue backlog, merge a few PRs and help get a new release out supporting django 3.2?

ad-m:
[...]

PR #816, opened 2024-03-27
Add support for newer Django versions, up to 5.0

ZbigniewRA:
This PR builds on top of PR #800, which adds support for Django 4.1, 4.2, and Python 3.11.

My additions:
- Added support for newer Django versions, up to 5.0.
- Added support for Python 3.12.
- SHA1PasswordHasher is removed since Django 5.1, so was removed from the test application.


johnthagen:
@ad-m would you mind reviewing this PR?

GabrielCastelo-31:
Will the Django-Guardian project still be maintained? I think that this PR is really important to keep the project alive.

johnthagen:
@GabrielCastelo-31 See

- #603

There has been some recent movement to give new maintainers permissions to move the project forward.

GabrielCastelo-31 (approved):
LGTM :+1:

.github/workflows/tests.yml
```diff
@@ -10,37 +10,39 @@ on:
...
-        include:
-          - python-version: "3.10"
-            django: 3.2.0
-          - python-version: "3.8"
-            django: 4.0.0
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        django: ["3.2", "4.1", "4.2", "5.0", "main"]
+        exclude:
+          - python-version: "3.12"
+            django: "4.2" # Python 3.12 is supported since Django 4.2.8.
```
thclark, line 21: Surely the `4.2` tag installs django `^4.2`, which would be `4.2.12` as of the time of writing. So I think we don't need this exclusion?

thclark (approved):
This is a really great PR, loads of attention to detail has been paid. Thanks so much for your work, folks!

johnthagen:
@cmaggiulli This would be a great PR to review.

cmaggiulli:
> a
Excellent will do

cmaggiulli:
@johnthagen excellent will do.  I will add my review by earlier next week. At a high level - are we all OK with dropping support for the versions of Django and python outlined in the MR?

johnthagen:
> Is the thought that this will be coupled with other MRs/features and eventually deployed as a new major release?

@cmaggiulli My vote is, yes. It's important to `django-guardian`'s long term health to add support for new Python and Django versions and drop support for end of life versions.

The fact there has not been a release in 3 years is another threat to its long term health because it will dissuade any new users from taking up the project.

In my opinion, the next short term goal should be releasing a new version with this MR and maybe other small MRs just to signal that it is alive again.

thclark (approved):

cmaggiulli:
This hasnt fallen off my radar - Im just running a day or two behind

example_project/settings.py
```diff
@@ -76,7 +76,6 @@
 
 PASSWORD_HASHERS = (
     'django.contrib.auth.hashers.MD5PasswordHasher',
```
cmaggiulli, line 78: I realize this is mostly irrelevant since this is an example app, and this MR does not add this hasher, but I would suggest we change this to use PBKDF2PasswordHasher which has been available from 3.2 onward.  Django lists this as not recommended as far back as 3.2 [not recommended](https://github.com/django/django/blob/c3d3af8ea369e16263c8330b1745ba101d1c8d62/django/contrib/auth/hashers.py#L703)

guardian/testapp/migrations/0006_auto_20230727_0658.py
```diff
@@ -0,0 +1,68 @@
+# Generated by Django 3.2.20 on 2023-07-27 06:58
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
cmaggiulli, line 6: At some point we should consider using a mix of both UUID primary keys and BigInt ( any non-char pks ) as we have experienced issues both in django proper and guardian upgrades with UUID fields.

example_project/requirements.txt
```diff
@@ -1,5 +1,5 @@
-Django~=2.2.9
-django-environ==0.4.5
-django-rosetta==0.9.3
+Django~=3.2.20
```
cmaggiulli, line 1: should we =3.2.20,<6

setup.py
```diff
@@ -28,30 +28,28 @@
     ],
     include_package_data=True,
     license='BSD',
-    install_requires=["Django>=2.2"],
+    install_requires=["Django>=3.2"],
```
cmaggiulli, line 31: may want to consider `install_requires=["Django>=3.2,<=5.0"]`

setup.py
```diff
@@ -12,7 +12,7 @@
 setup(
     name='django-guardian',
     version=version,
-    python_requires='>=3.5',
+    python_requires='>=3.8',
```
cmaggiulli, line 15: may want to consider `python_requires='>=3.8,<=3.12'`

setup.py
```diff
@@ -28,30 +28,28 @@
...
                  'Programming Language :: Python',
                  'Topic :: Security',
                  'Programming Language :: Python :: 3',
                  'Programming Language :: Python :: 3 :: Only',
-                 'Programming Language :: Python :: 3.5',
-                 'Programming Language :: Python :: 3.6',
-                 'Programming Language :: Python :: 3.7',
                  'Programming Language :: Python :: 3.8',
                  'Programming Language :: Python :: 3.9',
                  'Programming Language :: Python :: 3.9',
```
cmaggiulli, line 49: 3.9 listed twice

tox.ini
```diff
@@ -1,30 +1,28 @@
 [tox]
 downloadcache = {toxworkdir}/cache/
```
cmaggiulli, line 2: downloadcache is entirely deprecated and has no effect.  I propose we remove

tox.ini
```diff
@@ -1,30 +1,28 @@
...
-    {core,example,docs}-py{36,37,38}-django30,
-    {core,example,docs}-py{36,37,38,39}-django31,
-    {core,example,docs}-py{36,37,38,39,310}-django32,
-    {core,example,docs}-py{38,39,310}-django40,
-    {core,example,docs}-py{38,39,310}-djangomain,
+    {core,example,docs}-py{38,39,310}-django32,
+    {core,example,docs}-py{38,39,310,311}-django41,
+    {core,example,docs}-py{38,39,310,311}-django42,
+    {core,example,docs}-py{310,311,312}-django50,
+    {core,example,docs}-py{310,311,312}-djangomain,
```
cmaggiulli, line 8: My 0.02 but unless we intent on regularly testing against django's devel Im not sure if I would keep `{core,example,docs}-py{310,311,312}-djangomain`.  Not now - just a thought going forward.  I realize this is out of scope for this MR

cmaggiulli (approved):
Looks good to me.  I would like to add some test cases for uuid fields and a mix of direct and generic foreign key tables.  That was my biggest concern here.  I played around with the code a bit in our environment and it seemed to work.  Building those tests were more time consuming that I originally anticipated so Ill add afterwards.  All my comments are suggestions - nothing impactful

example_project/requirements.txt
johnthagen: @cmaggiulli In general capping versions is often not recommended as it removes the end consumer from being able to, for example, use an older django gaurdian that works fine with new Django but has not yet been tested at the time.

I think the general consensus is that it's better to leave these open and let the end consumer pin upper bounds in applications if needed but leave libraries more permissive. Especially for a library like Django that has stable backwards compatibility story.

setup.py
johnthagen: Same thought in https://github.com/django-guardian/django-guardian/pull/816#discussion_r1682714364

I would avoid upper bounds.

setup.py
johnthagen: For similar reasons to https://github.com/django-guardian/django-guardian/pull/816#discussion_r1682714364, I would recommend that libraries like this not cap and upper bound on Python versions or dependencies (unless absolutely necessary).

`django-guardian` is a pure Python package with no binary wheels, so I think it's even less useful to pin upper Python bounds.

cmaggiulli: Good to know - I actually didn't know that guardian was not packaged and distributed as a whl.  Thanks for the info!

johnthagen: It's packaged as a wheel, it's just a pure Python wheel (rather than any native complied code), thus should work fine on newer versions of Python and shouldn't need an upper bound

- https://pypi.org/project/django-guardian/#files

example_project/settings.py
ZbigniewRA: Changed.

guardian/testapp/migrations/0006_auto_20230727_0658.py
ZbigniewRA: I'm not sure what this is about, but I understand this is for later. Leaving it without change.

example_project/requirements.txt
ZbigniewRA: Leaving it as is.

setup.py
ZbigniewRA: Leaving it as is.

tox.ini
ZbigniewRA: Leaving it as is.

setup.py
ZbigniewRA: Fixed.

tox.ini
ZbigniewRA: Removed.

setup.py
ZbigniewRA: Leaving it as is.

thclark (approved):
Minor changes since my last review, happy

GabrielCastelo-31 (approved):

.github/workflows/tests.yml
```diff
@@ -10,37 +10,39 @@ on:
 
 jobs:
   tests:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
johnthagen, line 13:
```suggestion
    runs-on: ubuntu-24.04
```

So that it is pinned/reproducible?

.github/workflows/tests.yml
```diff
@@ -10,37 +10,39 @@ on:
...
-          - python-version: "3.10"
+          - python-version: "3.8"
+            django: 5.0
+          - python-version: "3.8"
             django: main
 
     services:
       mariadb:
-        image: mariadb:10.2
+        image: mariadb:latest
```
johnthagen, line 39: I suggest we pin these to latest version instead of `latest` since then people can't go back in time and reproduce the same environment.

.github/workflows/tests.yml
```diff
@@ -10,37 +10,39 @@ on:
...
         env:
           MYSQL_ROOT_PASSWORD: mariadb
         ports:
           - 3306:3306
       postgres:
-        image: postgres:10-alpine
+        image: postgres:latest
```
johnthagen, line 45: I suggest we pin these to latest version instead of `latest` since then people can't go back in time and reproduce the same environment.

.github/workflows/tests.yml
ZbigniewRA: Pinned to `ubuntu-24.04`,

.github/workflows/tests.yml
ZbigniewRA: Pinned to `mariadb:11.4.2`.

.github/workflows/tests.yml
ZbigniewRA: Pinned to `postgres:16.3`.

ZbigniewRA:
Tests I've run succeeded: [see here](https://github.com/RowAnalytics/django-guardian/actions/runs/10211907047)
But tests triggered by @thclark 15 hours ago failed ([see here](https://github.com/django-guardian/django-guardian/actions/runs/10159264626/job/28226071711)) with error: `operator does not exist: character varying = integer`
This is a mystery to me. Any ideas what happened?

BonaFideIT:
Hi,

We would also like to express our interest in continuing the maintenance of django-guardian.

We also think that an upgrade to the current LTS version of django (4.2) makes a lot of sense and should be a top priority.

Since the drop of support for end of life versions has already been briefly discussed here, we would like to remind you that all versions below 4.2 are end of life, see https://www.djangoproject.com/download/#supported-versions.

So we would like to suggest that only versions >=4.2 should be supported and the others <4.2 should be dropped. The simplest thing would be to support only the supported LTS versions, which should result in even less maintenance effort.

Drop Support Suggestion:
- django 3.2
- django 4.1

Support Suggestion:
- django 4.2 (LTS) + python 3.8, 3.9, 3.10, 3.11, 3.12 (>=4.2.8)
- django 5.0 + python 3.10, 3.11, 3.12
- django 5.1 + python 3.10, 3.11, 3.12

https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django

StephenChan:
I'm interested in seeing this PR merged as well.

@ZbigniewRA Pardon the mention, but I spent some time looking at those two CI runs you linked, and I noticed that the commit-trees used by each run (660125e failing, 55ca348 passing) have differences in shortcuts.py: https://github.com/django-guardian/django-guardian/compare/660125e..55ca348 

By using this PR's latest commit as a base and just applying those shortcuts.py differences, I replicated the failed run's errors: https://github.com/StephenChan/django-guardian/pull/2

The shortcuts.py differences don't seem to be related to this PR. Maybe they were from a different PR? In any case, it doesn't seem to be a concern for this PR's mergeability.

BonaFideIT:
@StephenChan These changes to shortcuts.py are from https://github.com/django-guardian/django-guardian/pull/768.

BonaFideIT:
You're right, I overlooked the changes from #812. These changes seem to be the problem why the tests are failing: https://github.com/django-guardian/django-guardian/actions/runs/9969776605/

BonaFideIT:
We fixed the tests in https://github.com/django-guardian/django-guardian/pull/827.

BonaFideIT (changes requested):
Looks good, but we would like to add changes from https://github.com/RowAnalytics/django-guardian/pull/3.

example_project/requirements.txt
michael-k: `~=` already includes an upper bound.  Also: This an example project, not django-guardian's requirements.

.github/workflows/tests.yml
```diff
@@ -10,33 +10,39 @@ on:
...
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
-        django: [2.2.17, 3.0.11, 3.1.3, 3.2.0]
-        include:
-          - python-version: "3.10"
-            django: 3.2.0
-          - python-version: "3.8"
-            django: 4.0.0
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        django: ["3.2", "4.1", "4.2", "5.0", "main"]
```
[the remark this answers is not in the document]
ZbigniewRA: I don't know how it works. If you'll let me know how the `test.yml` file should look now I will fix it.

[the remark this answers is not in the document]
BonaFideIT: We run the checks here: https://github.com/BonaFideIT/django-guardian/pull/5 and an it worked. @michael-k what exactly is the problem here?

BonaFideIT (approved):

.github/workflows/tests.yml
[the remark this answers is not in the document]
BonaFideIT: I triggered the github checks and all checks have passed.

BonaFideIT:
@brianmay and @thclark Can you please review this PR again?

.github/workflows/tests.yml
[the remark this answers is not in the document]
michael-k: The tests won't fail, but they'll install the wrong version.  `Django~=4.1` means `Django>4.1,==4.*`, see https://peps.python.org/pep-0440/#compatible-release

Example: https://github.com/BonaFideIT/django-guardian/actions/runs/11381238190/job/31662202522?pr=5#step:6:54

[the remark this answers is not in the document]
michael-k:
```suggestion
        django: ["3.2.25", "4.1.13", "4.2.16", "5.0.9", "main"]
```

.github/workflows/tests.yml
```diff
@@ -10,33 +10,39 @@ on:
...
+        exclude:
+          - python-version: "3.12"
+            django: "4.2" # Python 3.12 is supported since Django 4.2.8.
```
michael-k, line 21: No need to exclude it.  Even if you put `4.2.0` in the matrix, with `~=` it won't install `4.2.[0-7]`.

```suggestion
```

.github/workflows/tests.yml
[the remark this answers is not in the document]
thclark: I agree with this I think it's the last tweak before we merge. @ZbigniewRA if you could make this change please? Then we'll get this merged and make a release!!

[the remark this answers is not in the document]
BonaFideIT: @michael-k Thanks for clarifying! Now we understand the problem and we completely agree with you.

.github/workflows/tests.yml
ZbigniewRA: Removed as suggested.

.github/workflows/tests.yml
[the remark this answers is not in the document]
ZbigniewRA: Changes to versions suggested by michael-k.

thclark (approved):

thclark:
Given that this has been re-approved ad-infinitum and the only changes made since previous reviews have been to address their comments, I'm comfortable that there's enough consensus now to merge this. Thank you so much for your work everybody!

PR #825, opened 2024-09-01
Add GuardianGroupMixin to support custom group models

BonaFideIT:
Hi,

since we like django-guardian, we want to contribute to it.

We would like to add the possibility to use a mixin instead of monkey patch the group model.

With this change it should be easy to use a custom group model like the following:

mixins.py
```python

class CustomGroupPermissionsMixin(PermissionsMixin):
    groups = models.ManyToManyField(
        "CustomGroup",
        verbose_name=_("groups"),
        blank=True,
        help_text=_(
            "The groups this user belongs to. A user will get all permissions "
            "granted to each of their groups."
        ),
        related_name="user_set",
        related_query_name="user",
    )

    class Meta(PermissionsMixin.Meta):
        abstract = True

```

models.py
```python

class CustomGroup(Group, GuardianGroupMixin):
    foo = models.CharField(max_length=120)
    ...


class User(AbstractBaseUser, CustomGroupPermissionsMixin, GuardianUserMixin):
    email = models.EmailField(
        verbose_name=_("email address"), max_length=200, unique=True, db_index=True
    )
    ...


class CustomGroupObjectPermission(GroupObjectPermissionAbstract):
    group = models.ForeignKey(CustomGroup, on_delete=models.CASCADE)

    class Meta(GroupObjectPermissionAbstract.Meta):
        abstract = False

```

settings.py
```python
...
GUARDIAN_MONKEY_PATCH_USER = False
GUARDIAN_MONKEY_PATCH_GROUP = False
GUARDIAN_GROUP_OBJ_PERMS_MODEL = 'myapp.CustomGroupObjectPermission'
...
```

~~Additional we add a check for the User model to check if `GUARDIAN_MONKEY_PATCH` was used correctly.~~


BonaFideIT:
We have fixed some bugs and added the documentation for the changes in the userguide.

BonaFideIT:
This solutions has the same problem as mentioned in https://github.com/django-guardian/django-guardian/pull/504#issuecomment-1472580496.

Currently our solution is to use a custom `ModelBackend` (see https://docs.djangoproject.com/en/4.2/topics/auth/customizing/):

backends.py
```python
class CustomModelBackend(ModelBackend):

    def _get_group_permissions(self, user_obj):
        return Permission.objects.filter(group__in=user_obj.groups.all())

```

settings.py
```python

AUTHENTICATION_BACKENDS = (
    "myapp.backends.CustomModelBackend",
)

```

We have opened a ticket (https://code.djangoproject.com/ticket/35792) and a PR https://github.com/django/django/pull/18633.

BonaFideIT:
Good news, out PR https://github.com/django/django/pull/18633 was merged into django.

Once included in a django release, django-guardian can support custom group models without overwriting the ModelBackend.

thclark:
@BonaFideIT looks promising (and frustratingly too late for me, as I'm currently in the middle of a beast of a migration that this would have solved).

I've converted it to a draft though, as it's showing as requiring review and (correct me if I'm wrong) but it doesn;t seem like we're there yet.

BonaFideIT:
@thclark The failing tests have been merged into devel. We just need to merge devel into this draft and then this should be ready. We have this change already in production (while converting an existing system, this is how this came to be) for two weeks without any issues yet.

BonaFideIT:
We have added the changes that fixed the tests.

BonaFideIT:
We have added a sample project so that our changes can also be tested. We found a few more errors and fixed them immediately.

guardian/conf/settings.py
```diff
@@ -20,7 +20,15 @@
 GET_INIT_ANONYMOUS_USER = getattr(settings, 'GUARDIAN_GET_INIT_ANONYMOUS_USER',
                                   'guardian.management.get_init_anonymous_user')
 
-MONKEY_PATCH = getattr(settings, 'GUARDIAN_MONKEY_PATCH', True)
+MONKEY_PATCH_USER = getattr(settings, "GUARDIAN_MONKEY_PATCH_USER", True)
```
thclark, line 23: I think this is the right way to go, and we can put this into the v3 release, but we should have a deprecation warning so if `GUARDIAN_MONKEY_PATCH` is present we barf and tell the user to rename it to `GUARDIAN_MONKEY_PATCH_USER`.

guardian/core.py
```diff
@@ -106,7 +103,10 @@ def get_user_filters(self, obj):
         model = get_user_obj_perms_model(obj)
         related_name = model.permission.field.related_query_name()
 
-        user_filters = {'%s__user' % related_name: self.user}
+        # generate filters
+        user_filters = {f'{related_name}__user': self.user}
+
+        # check for generic model
```
thclark, line 109: comments are self evident

thclark (changes requested):
Thanks for this @BonaFideIT - just looking in detail for PRs I'd like to get into `v3` and this is definitely one of them. I had a coupe of very trivial comments if you wouldn't mind tweaking I can merge away.

guardian/core.py
BonaFideIT: Yes, that makes sense. The comments have been removed.

guardian/conf/settings.py
BonaFideIT: Yes, that makes sense, too. `ImproperlyConfigured` will be raised with the hint, that the settings is deprecated now.

BonaFideIT:
We added the suggested changes @thclark, please check it again :)

thclark:
Brilliant, thanks so much @BonaFideIT

PR #838, opened 2024-12-28
get_groups_with_perms bug

okaluk:
Hello,

in a project im working with, i discovered a bug.

When overriding the guardian-setting GROUP_OBJ_PERMS_MODEL with a custom model, we use the generic GroupObjectPermissionAbstract mixin to implement the custom model.

We use the get_groups_with_perms and when we want to get the perms attached, it crashes.

When investigating this crash, i discovered our custom model is checked against "is GroupObjectPermission", which cannot be true even if it should be true, because we dont have the content_object_id key inside our model. 

I build a testcase to show this bug in a pretty basic way. I already have the fix for this bug.

~~It would be nice, to run the CI to show that the bug exists, i would then push my fix to validate it fixes the testcases~~

Best Wishes


guardian/testapp/testsettings.py
```diff
@@ -5,7 +5,7 @@
 
 env = environ.Env()
 
-DEBUG = False
+DEBUG = True
```
okaluk, line 8:
```suggestion
DEBUG = False
```
This is not intended and has to be reverted

okaluk:
i created a PR in my branch to run the pipeline:

Failing Tests:
https://github.com/okaluk/django-guardian/pull/2/checks?check_run_id=34938165970

Fixed Tests:
https://github.com/okaluk/django-guardian/pull/2/checks?check_run_id=34938201319

BonaFideIT (approved):
Thank you for finding the bug! Looks good for us.

BonaFideIT:
@thclark @brianmay please review this PR.

brianmay (approved):
I think I am not the right person to be reviewing this. But I don't see anything obviously wrong, so will approve.

BonaFideIT (approved):

thclark:
> > I think I am not the right person to be reviewing this. But I don't see anything obviously wrong, so will approve.
> 
> Is it possible to restore our write access? our second review does not seem to count, I guess this is due to not having sufficient rights?

I'll have a look!

thclark:
> > I think I am not the right person to be reviewing this. But I don't see anything obviously wrong, so will approve.
> 
> Is it possible to restore our write access? our second review does not seem to count, I guess this is due to not having sufficient rights?

Thanks @BonaFideIT - I think your access was reduced because you didn't have 2FA set up previously, and I overhauled the security policy to require it. I see that's no longer a problem so triggered an invitation to the reviewers team.

guardian/testapp/migrations/0007_genericgroupobjectpermission.py
```diff
@@ -0,0 +1,31 @@
...
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('auth', '0012_alter_user_first_name_max_length'),
+        ('contenttypes', '0002_remove_content_type_name'),
+        ('testapp', '0006_auto_20230727_0658'),
```
thclark, line 12: I believe this will break everybody's migrations as `testapp` isn't a django app (something on your machine? Or the example app?)

thclark (changes requested):
~~This will break every user's CI on migration, leading to one hell of a mess, because it's dependent on a non-core app (@BonaFideIT even though I've added you please don't approve this!).~~

~~@okaluk Please can you outline in more detail the need for a migration and what you're trying to do here? Which issue does this PR address? If it's not against an issue, please can you raise an issue so we take it a step at a time to determine what problem this is solving and find a solution that won't break everyone's installations?~~

guardian/testapp/migrations/0007_genericgroupobjectpermission.py
BonaFideIT: not sure what you are referring to. This app it the test app for running the tests. 

How will that break the CI? 

The testcases have all passed in this pull request? 

This migration depends on two standard django apps (contrib auth and contrib contenttypes) and on the "testapp" which is just the migration in the same app that already existed.

thclark: @BonaFideIT I'm looking to find where testapp is imported, I think it'd be highly unusual to be adding an extra app into peoples production databases just for us to test guardian in our dev/CI. As far as I know this app shouldn't actually be registered in a production environment. If it *is*, then this PR might be OK but then that's a whole different architectural can of worms.

thclark: I inspected my production database. The `testapp` app isn't registered and none of its tables appear, so I'm asusming migrations that depend on it being registered will fail. I'll recreate the error that I expect to get on migration.

BonaFideIT: This is not a production issue. The app is used in the CI. The migrations is also just used in the CI. The part of code you quoted will never be executed in production. :) Thanks for looking into it.

thclark: @BonaFideIT OMG You're right, I'm so sorry it's a migration in testapp itself!

OK no problem I'll actually merge this then. It'll make it into `3.0.0rc2`.

thclark (approved):

guardian/testapp/migrations/0007_genericgroupobjectpermission.py
BonaFideIT: cheers, thanks :)

PR #841, opened 2025-01-24
Release 3.0.0

thclark:
## Status Note

Release candidate 3.0.0rc1 is PUBLISHED! https://pypi.org/project/django-guardian/3.0.0rc1/

**ALL - Please install and try it, let us know if there are any problems**

## This release

This is the culmination of more than two years of commits by various dedicated people, and a lot of wrangling to resurrect `django-guardian` from the dead. It aims to bring guardian up to date with a new team of maintainers and hit reset on the project going forward - it also encompasses updates to be compatible with the latest, greatest versions of django.

## Why a Major version?

We're reasonably sure that this major version does not actually constitute any breaking changes. However, there's been a complete turnover of the maintenance team, and we're releasing some features which we didn't touch, review or participate in in any way (although all commits have been doubly reviewed when they were merged).

I've gone over to review this entire release as best as I can but without many days of deep introspection, it's realistically possible to miss some changes that have implications deemed "breaking".

Sensible app developers should be pinning their dependencies like >2, <3 (or tighter) so in most cases, the update to this version (v3) will be a conscious choice as opposed to automatic update. It's therefore much safer for us to make a major version release (then possibly have to update the release notes!), than it is to do a minor version release which risks breaking people's systems.

To further mitigate this risk we'll first issue Release Candidate versions which people can install and give feedback on.

In the past, (eg the 2.0.0 release), removal of old django versions (which we have done in this release) was seen to be a breaking change. With modern dependency resolvers like [uv](https://github.com/astral-sh/uv) and [poetry](https://python-poetry.org/) this should no longer break people's installs (simply prevent them from using newer versions) so in future this will not be deemed a breaking change.


BonaFideIT:
Thanks for all the changes, we will try the current state in our environments to see if it breaks something.

What about removing the support for django 3.2 because it is not longer supported:

    Hi,

    We would also like to express our interest in continuing the maintenance of django-guardian.

    We also think that an upgrade to the current LTS version of django (4.2) makes a lot of sense and should be a top priority.

    Since the drop of support for end of life versions has already been briefly discussed here, we would like to remind you that all versions below 4.2 are end of life, see https://www.djangoproject.com/download/#supported-versions.

    So we would like to suggest that only versions >=4.2 should be supported and the others <4.2 should be dropped. The simplest thing would be to support only the supported LTS versions, which should result in even less maintenance effort.

    Drop Support Suggestion:
    - django 3.2
    - django 4.1

    Support Suggestion:
    - django 4.2 (LTS) + python 3.8, 3.9, 3.10, 3.11, 3.12 (>=4.2.8)
    - django 5.0 + python 3.10, 3.11, 3.12
    - django 5.1 + python 3.10, 3.11, 3.12

    https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django

_Originally posted by @BonaFideIT in https://github.com/django-guardian/django-guardian/issues/816#issuecomment-2323319279_

BonaFideIT:
@thclark fyi we have run our CI with the new release candidate and none of our tests failed so far. Looks like we have a promising candidate for a new stable release. (^_^)/

BonaFideIT:
@thclark Can you please create a new rc release with the current devel branch? Some of the merged changes from devel are not in the current rc release.

johnthagen:
@thclark I ran `3.0.0rc1` through our unit test suite and everything passed 🚀 .

I agree with @BonaFideIT we should ideally push out one final `rc2` and then shortly after publish `3.0.0` since I think further delays are probably more harmful to `django-guardian`'s image than simply getting it out.

tidenhub:
@thclark I tested `3.0.0rc1` today as well. Our test suite didn't show any errors with Django 5.2.

thclark:
Having had a few people try this out with no issues, I'm comfortable... ish!

Five years of commits is a lot of code but here goes... LET'S RELEASE IT, BABY!

johnthagen:
@thclark Thank you! You are our true guardian.

PR #844, opened 2025-01-25
Support Python 3.13 and Django 5.1

johnthagen:
- Add support for Python 3.13
- Add support for Django 5.1
- Add CI matrix based  on supported Python versions for Django
  - https://docs.djangoproject.com/en/5.1/faq/install/#what-python-version-can-i-use-with-django


johnthagen:
CC @thclark. Please squash when merging this, thanks.

thclark (approved):

PR #845, opened 2025-01-25
Replace Restructured Text with Markdown

### LINKED ISSUE #828: Replace restructuredText with markDown
## Feature request

We should consider an upgrade in the docs to allow easier contributions.

### Background

The current documentation is built in restructuredText, which most people seem to find quite painful to write compared to markdown. Since the structure of the docs isn't particularly complicated, I think it'd be a good thing to switch to a markdown-based system like [mkdocs-material](https://squidfunk.github.io/mkdocs-material/) which seems to be seeing a lot of success (eg in projects like [strawberry](https://strawberry.rocks/docs/integrations/django)).

### Hosting

Whether to continue publishing on readthedocs or to go for the simpler github-pages I don't mind, although think it's quite useful to keep all the legacy documentation against historical versions, so would be inclined to update the readthedocs build rather than burn the building down in that respect :)

### Proposed solution 
- [ ] Redraft docs into markdown
- [ ] Render with material-mkdocs instead of sphinx
- [ ] Reinstate build pipeline on readthedocs

I'm willing to take this on myself but any assistance is extremely welcome.

dpgraham4401:
## Description

This PR migrates the documentation from [reStructured Text](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) files to [Markdown](https://www.markdownguide.org/) files, as well as a [Sphinx](https://www.sphinx-doc.org/en/master/index.html#) build to a [MkDocs](https://www.mkdocs.org/) build tool. The new MkDocs based documentaion makes use of the `mkdocs-material` theme, and also has everything needed (javascript and overrides) needed to operate on [readthedocs](https://about.readthedocs.com/), including server-side searching and a documentation version dropdown. 

I may have missed some artifacts created during the conversion from .rst to .md files. The .rst files were initially converted to markdown using the [pandoc tool](https://pandoc.org/), then manually cleaned up. There were a lot of little artifacts, I didn't realize how long this process was going to take.

### Updates to the docs content

1. This PR migrates the docstring style from Sphinx to [google's style](https://google.github.io/styleguide/pyguide.html) ([examples](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)). I migrated the sphinx based docstrings to google because I could never get the mkdocstring Sphinx parser to work with Guardian's existing docstring 100%. Google is mkdocstring default, and i believe it's generally less restrictive than Numpy's docstring style. However, there's nothing stopping us from adopting Numpy docstring styles.  

2. Adds a logo, a super simple shield, see below image. I'm not actually tied to this, and it needs to be aligned better, if it was merged. I was just sick of looking at the default logo MkDocs uses. 
3. There are some minor re-wordings here and there. 

## Links

closes #828
[mkdocs documentation](https://squidfunk.github.io/mkdocs-material/)
[mkdocstring](https://mkdocstrings.github.io/usage/)
[Python version status](https://devguide.python.org/versions/)
[My dummy project showing a working example of MkDocs deployed to read the docs with functioning ReadTheDocs-power-search and version fly out](https://hazsystems.readthedocs.io/en/latest/)

image


dpgraham4401:
@thclark checking that this is close to what you envisioned; I'm open to suggestions.

It's not 100% ready. The API Reference section is incomplete. I likely need to modify the docstrings of Guardians API, and employ dependency like mkdocstring.

I'd also like to make sure that there's a step in the CI to check that the documentation builds successfully before creating a release. Haven't looked at what's in there.

dpgraham4401:
I see that in the 3.0 release candidate that we're still supporting Python 3.8 and Django 3.something. I can revert/update to reflect what's in there.

dpgraham4401:
Note: we can get rid of the `overrides` directory if we don't care about adding a dropdown in the header with a version dropdown (see images from my dummy repo). The black fly out version selector will automatically be added by readthedocs. 

image

image

thclark:
Oh my goddddd  🤩

thclark:
Looking so cool @dpgraham4401!

thclark:
> @thclark checking that this is close to what you envisioned; I'm open to suggestions.
> 
> It's not 100% ready. The API Reference section is incomplete. I likely need to modify the docstrings of Guardians API, and employ dependency like mkdocstring.
> 
> I'd also like to make sure that there's a step in the CI to check that the documentation builds successfully before creating a release. Haven't looked at what's in there.

Still, great progress though!

CI cis best done through integration with RTD, but I'm struggling to make it play ball correctly so it would definitely help to have a step doing this.

thclark:
> I see that in the 3.0 release candidate #841 that we're still supporting Python 3.8 and Django 3.something. I can revert/update to reflect what's in there.

python 3.8 and django 3.2 are now EOL so I've no problem removing that support and can do so in a separate PR; don't worry about making the docs backward compatible.

requirements.txt
```diff
@@ -2,4 +2,3 @@ Django>=4.2
 django-environ
 bumpversion
 mock
-sphinx_rtd_theme==3.0.2
```
thclark, line 5: Is there a dependency we now need to have for mkdocs? eg to preview builds in our local tools?

thclark:
> Note: we can get rid of the `overrides` directory if we don't care about adding a dropdown in the header with a version dropdown (see images from my dummy repo). The black fly out version selector in the bottom right will automatically be added by readthedocs.

I think it looks smart the way you have it, but I guess we may need to maintain that javascript over time. Your call as to which way to go - you're the author :)

thclark:
@dpgraham4401 would it be helpful for you to be added to the readthedocs maintainers so you can see the build processes on this branch? If so let me know your email (send it to guardian@octue.com which is me, if you don't want to publicly put it online).

dpgraham4401:
> I think it looks smart the way you have it, but I guess we may need to maintain that javascript over time. Your call as to which way to go - you're the author :)

I wish I could take credit for the javascript, but it's straight out of the [readthedoc's documentation on using mkdocs](https://docs.readthedocs.com/platform/stable/intro/mkdocs.html) so the only maintainence should be minimal I think (I also put the link to that documentation in the javascript). 

> would it be helpful for you to be added to the readthedocs maintainers so you can see the build processes on this branch? If so let me know your email (send it to [guardian@octue.com](mailto:guardian@octue.com) which is me, if you don't want to publicly put it online).

That might make things easier while getting this up and running, I'll email you. Thanks!

dpgraham4401:
@thclark You can [see the draft documentation here](https://django-guardian--845.org.readthedocs.build/en/845/). 

I think it's generally ready.

thclark (approved):
Stunning work @dpgraham4401 thank you very much!

dpgraham4401:
Already found some typos and I need to double check the mixins API reference.

dpgraham4401:
@thclark I just found 1 thing in the `PermissionListMixin` that we probably want changed before the 3.0 release. It's technically still backward compatible, but the documentated public API would change.

There was a typo in the method name `get_get_objects_for_user_kwargs` (it has an extra `"get_"`).

The latest commit in this PR creates a new method called `get_objects_for_user_kwargs` with the same implementation. The old `get_get_objects_for_user_kwargs` just forwards calls to this method. The new documentation only shows `get_objects_for_user_kwargs` in the API Reference.

Let me know what you think. Happy to roll back these changes to keep this PR more focused.

I don't see any tests that directly use this method. 

[Old method documentation for `get_get_objects_for_user_kwargs`](https://django-guardian.readthedocs.io/en/stable/api/guardian.mixins.html#guardian.mixins.PermissionListMixin.get_get_objects_for_user_kwargs)

thclark:
> @thclark I just found 1 thing in the `PermissionListMixin` that we probably want changed before the 3.0 release. It's technically still backward compatible, but the documentated public API would change.
> 
> There was a typo in the method name `get_get_objects_for_user_kwargs` (it has an extra `"get_"`).
> 
> The latest commit in this PR creates a new method called `get_objects_for_user_kwargs` with the same implementation. The old `get_get_objects_for_user_kwargs` just forwards calls to this method. The new documentation only shows `get_objects_for_user_kwargs` in the API Reference.
> 
> Let me know what you think. Happy to roll back these changes to keep this PR more focused.
> 

Although utterly terribly named, I think it *technically* isn't a typo. It's the function that `get`s the `kwargs` that you supply to `get_objects_for_user`, as in `get(get_objects_for_user)kwargs`.

If we were going to remove it, I'd rather deprecate that as you just know someone'll come out of the woodwork having overridden it and we'll break their use case. I'd rather get into things like clearing that up on v4 than v3.

So if you could roll back that one it'd be great, thanks!

dpgraham4401:
> f we were going to remove it, I'd rather [deprecate](https://pypi.org/project/Deprecated/) it over a couple of releases, as you just know someone'll come out of the woodwork having overridden it and we'll break their use case. I'd rather get into things like clearing that up on v4 than v3.

@thclark that makes perfect sense, I kinda jumped the gun. easy revert. Will do that in an hour.

PR #849, opened 2025-01-31
Fix readthedocs search text contrast

dpgraham4401:
# Description 

Small fix to make the sure search text is always readable in dark mode. 

- [readthedocs addon css variable reference](https://docs.readthedocs.com/platform/stable/addons.html\#css-variables-reference)
- [Adding extra css to mkdocs](https://www.mkdocs.org/user-guide/configuration/#extra_css)

This should fix this cosmetic/accessibility issue with the latest version of the docs. 

image

### DIFF
--- a/docs/css/readthedocs.css
+++ b/docs/css/readthedocs.css
@@ -1,0 +1,4 @@
+:root {
+    /* Ensure that search text color is not low contrast */
+    --readthedocs-search-color: black;
+}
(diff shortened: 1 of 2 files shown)


thclark (approved):

dpgraham4401:
@thclark those changes weren't ready yet, it's OK but the changes don't fix the problem like I initially thought they would. 

I should have noted that I opened the changes as draft PR so I could see the changes after readthedocs built them since we can't view the readthedocs search model locally.

PR #854, opened 2025-02-25
`get_users_with_perms`: filter out inactive users

### LINKED ISSUE #944: Filter out Inactive Users
Based on this PR: https://github.com/django-guardian/django-guardian/pull/854/files

> @BonaFideIT Although it is unlikely (and also illogical) but, someone may have disabled the “is_active” field in their project. Even though it improves performance, how about defining it as a configuration parameter?
> 
> For example, if there were a parameter like “GUARDIAN_WORK_ONLY_ACTIVE_USERS” with a True/False value, similar to “GUARDIAN_RAISE_403,” and Django Guardian operated based on this field, it might be safer, what do you think?

raphaelmerx:
Fix a small bug in `get_users_with_perms`, which would yield users that are not active, as opposed to behaviour in `user.has_perm`.

Also adds a regression test for this.


guardian/shortcuts.py
```diff
@@ -319,7 +319,7 @@ def get_users_with_perms(obj, attach_perms=False, with_superusers=False,
             qset = qset | Q(groups__in=group_ids)
         if with_superusers:
             qset = qset | Q(is_superuser=True)
-        return get_user_model().objects.filter(qset).distinct()
+        return get_user_model().objects.filter(qset).filter(is_active=True).distinct()
```
iamMHZ, line 322: To maintain backward compatibility, we may need to make it an input parameter of the `get_users_with_perms` function.

BonaFideIT: We agree, this would also break custom user model support that does not have this property.
Maybe its worth discussing removing the behaviour in has_perm instead. 

I propose we

* introduce a settings that allows activation of the behaviour 
* add a check if the user model has the property "is_active" 
* add a warning if the setting is used and is_active is not present.

including respective tests obviously. Any objections and maybe volunteers? ;)

Natgho:
@BonaFideIT  Although it is unlikely (and also illogical) but, someone may have disabled the “is_active” field in their project. Even though it improves performance, how about defining it as a configuration parameter?

For example, if there were a parameter like “GUARDIAN_WORK_ONLY_ACTIVE_USERS” with a True/False value, similar to “GUARDIAN_RAISE_403,” and Django Guardian operated based on this field, it might be safer, what do you think?

Natgho:
Hi @raphaelmerx Would you consider handling PR? This is a good idea; if you don't have the time, I can take responsibility for it.

raphaelmerx:
Sorry won't have the time, happy for you to take over this PR

thclark:
Closing in favour of #956 - discussion ongoing

PR #856, opened 2025-04-08
add support for non-User models in `get_identity` util

nikatlas:
This PR encourages the decoupling of the authorization from the authentication models used in Django.
In most applications the two are mixed together and are served by the same model, however in complex multi-tenant applications the authentication is strictly separated by the authorization. 

The aim of this PR is to enable `django-guardian` to work with separate permission models. 

## Changes - explained

with `get_user_obj_perms_model` we are allowed to set the permission model that will be used for persisting permissions in the database. this allows us to point the user FK to a non-User model, however, the `get_identity` still relies on the User model explicitly.

the only requirement for the permission model is to extend the `PermissionsMixin` from django, rather to be the `get_user_model`.

### DIFF
--- a/guardian/utils.py
+++ b/guardian/utils.py
@@ -83,6 +83,7 @@ def get_identity(identity):
         identity = get_anonymous_user()
 
     Group = get_group_obj_perms_model().group.field.related_model
+    UserModel = get_user_obj_perms_model().user.field.related_model
 
     # get identity from queryset model type
     if isinstance(identity, QuerySet):
@@ -93,12 +94,12 @@ def get_identity(identity):
             return None, identity
 
     # get identity from first element in list
-    if isinstance(identity, list) and isinstance(identity[0], get_user_model()):
+    if isinstance(identity, list) and isinstance(identity[0], UserModel):
         return identity, None
     if isinstance(identity, list) and isinstance(identity[0], Group):
         return None, identity
 
-    if isinstance(identity, get_user_model()):
+    if isinstance(identity, UserModel):
         return identity, None
     if isinstance(identity, Group):
         return None, identity


nikatlas:
I am adding this tiny PR as an initiator to the proposed change. 
Without having the full context of the code here, I will be expecting some feedback on the initiative. I can better prepare the PR if necessary. 

In my project, I actually monkey patched the code in a similar fashion and managed to make `django-tenant-users` work along with `django-guardian`. 

@lukaszb @brianmay

nikatlas:
@thclark @ad-m

dpgraham4401:
Hey @nikatlas, interesting idea. I've never used django for an instance where the authentication user model wasn't the same as the authorization user model. It's such a fundamental change for guardian, it might have to be implemented in a backward incompatible way. 

Seems like the changes in the PR would be backwards compatible as long as:
1. Projects are using the defatul `UserObjectPermission` model
2. projects that that override classes like `UserObjectPermissionBase` but don't override the `user` field

I believe the current implementation (relying on `related_model`) might have some difficulties with `GenericForiegnKeys`, but not sure if that's even a problem.

nikatlas:
I took inspiration from the `Group` model which uses the same logic to infer the proper model.
In the documentation, there is already the possibility to override both the default `UserObjectPermission` model ([here](https://django-guardian.readthedocs.io/en/stable/configuration.html#guardian-user-obj-perms-model)) and the default `GroupObjectPermission` model ([here](https://django-guardian.readthedocs.io/en/stable/configuration.html#guardian-group-obj-perms-model)). 

It seems changing the Group is a common use-case and thus the rest of the code accounts for it. (e.g. the `get_identity` otherwise it would break).

Considering that permission-wise the Group/User are both working along a similar interface, I do not see the permission-checking code breaking.

> It's such a fundamental change for guardian, it might have to be implemented in a backward incompatible way.
From my point of view, since this is an opt-in feature, it is a valid non-major update. 
 
> Seems like the changes in the PR would be backwards compatible as long as:
> 1. Projects are using the defatul `UserObjectPermission` model
> 2. projects that that override classes like `UserObjectPermissionBase` but don't override the `user` field

In my implementation, am actually overriding the `user` FK field actually to point to a different `authorization-only` model. However, that model still needs to proxy some fields from the `authentication` model. 
Specifically, the is_authenticated seems to be required for the ANONYMOUS_USER support [here](https://github.com/nikatlas/django-guardian/blob/8ef2fd2c8c4bc38c755633394499c3d7cb69247b/guardian/backends.py#L21-L42)

> I believe the current implementation (relying on `related_model`) might have some difficulties with `GenericForiegnKeys`, but not sure if that's even a problem.

Can you elaborate on potential issues here? The `user` field is not a GenericForeignKey as I understand, so that should be fine.

------------

I am going to outline some places that use the get_user_model with some comments:
- The `check_user_support` will receive the __authorization__ model and thus needs to proxy the `is_authenticated` property. Or we can add a `authentication_model` getter to make it more flexible. 
https://github.com/nikatlas/django-guardian/blob/8ef2fd2c8c4bc38c755633394499c3d7cb69247b/guardian/backends.py#L21-L42
- There is a seemingly deprecated functionality to patch the User model (its not documented on the latest version - seems to be 10 years old code) 
https://github.com/nikatlas/django-guardian/blob/d34748811b2b01150e71e53ab37209176b655f0b/guardian/__init__.py#L19-L31
- The admin.py file is using the `get_user_model`. Changes will be probably needed to fully support the __authorization__ model
https://github.com/nikatlas/django-guardian/blob/8ef2fd2c8c4bc38c755633394499c3d7cb69247b/guardian/admin.py#L7

These seem to be the core places that need to be touched for a full implementation.

In my opinion, `guardian` is a cool library but should not couple with __authentication__ at all since there is no __authentication__ feature implemented.

-----------

@vimpirate I would appreciate feedback on the above.

nikatlas:
I understand the concerns. 
Indeed the `UserObjectPermissionsBase` is hard pointing to the User and that can create problems in existing projects.

I will be testing this new approach in a new repo and keep updating this PR with issues we may find until its mature enough for another discussion.

thclark:
Hi @nikatlas thanks for the PR, is there an issue for this? There isn't one linked...

nikatlas:
@thclark No I outlined the proposal in the PR here directly.

nikatlas:
@thclark @dpgraham4401 I created an issue to track the progress of this idea. 

https://github.com/django-guardian/django-guardian/issues/861

PR #859, opened 2025-05-07
Fix formatting issues in README

johnthagen:
Most of these look like things missed when the README was converted from RST to Markdown.

### References

- https://github.com/django-guardian/django-guardian/pull/846/commits/864eaf266b34cb8d337c8efd85039c0f1f148ddf
- #846

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,8 +30,8 @@ We need to hook `django-guardian` into our project.
 
 ```python
 INSTALLED_APPS = (
-...
-'guardian',
+    ...
+    'guardian',
 )
 ```
 
--- a/README.md
+++ b/README.md
@@ -29,25 +29,25 @@ We need to hook `django-guardian` into our project.
 1. Put `guardian` into your `INSTALLED_APPS` at settings module:
 
 ```python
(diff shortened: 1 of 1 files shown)


johnthagen:
@thclark Please squash these commits when merging.

johnthagen:
@thclark Related question, should `main` be set as the repository's `default` branch instead of `next`? Otherwise, PR's will default target `next` instead of `main`. Is this intentional?

thclark:
@johnthagen 

> Please squash these commits when merging.

> PR's will default target next instead of main. Is this intentional?

Ideally, I'd like main to be stable (ie match the latest release) whilst `next` to be the default to check out and open a PR into. But it doesn't seem like this is possible, so I set `next` as default, which honestly I don't think helps anyone. Maybe I should just collapse the two and set `main` as default. Do you agree?

thclark (approved):

johnthagen:
> Maybe I should just collapse the two and set main as default. Do you agree?

I do agree. I think keeping the same branch structure as 99% of other projects will make contributing easier for newcomers.

PR #870, opened 2025-05-15
Remove unused `mock` install from PyPI

### LINKED ISSUE #852: Remove `mock` development dependency
I noticed a dependency on the `mock` PyPI package:

https://github.com/django-guardian/django-guardian/blob/e3dd3f3b47eebc1ef4a51f1ee327efad55d11b72/requirements.txt#L4

> mock is now part of the Python standard library, available as [unittest.mock](https://docs.python.org/dev/library/unittest.mock.html) in Python 3.3 onwards.

We should be able to drop this dependency since we support only versions of Python greater than 3.3.

johnthagen:
Closes #852

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,7 +130,6 @@ dev = [
     "mkdocs-material-extensions>=1.3.1",
     "mkdocstrings>=0.29.1",
     "mkdocstrings-python>=1.16.10",
-    "mock>=5.1.0",
     "mypy>=1.15.0",
     "mysqlclient>=2.2.7",
     "psycopg2-binary>=2.9.10",


johnthagen:
@thclark Side question, is there a reason that you gitignore `uv.lock`? I realize django-guardian is a package, but checking in `uv.lock` and updating it on some cadence would allow CI to be reproducible as well as local dev experience.

dpgraham4401 (approved):

dpgraham4401:
Some relevant info, astral-sh/uv#10730, I was surprised to find the astral team recommends adding the uv.lock file for libraries/packages as well as applications/projects.

thclark:
> astral-sh/uv#10730

They do make some good points, @dpgraham4401 by my main thing is the idea of not wanting to deal with endless merge conflicts in the lock file - so I'm a bit resistant to it. If we had better controls around requiring linear history the problem would reduce, but requiring more sophisticated git flows makes life tougher for contributors.

thclark (approved):

johnthagen:
> Thoughts? Perhaps there's a good strategy for this I'm not aware of?

I think in practice if all dependencies are simply updated every so often, other contributors don't need to update them and conflicts are not that often. I doubt that contributors will need to change the dependencies of this project often (though I recognize the irony that this PR does 😆).

I don't have a strong preference, but I think checking in the lockfile would be a net positive.

PR #871, opened 2025-05-15
Migrate Pytest config to pyproject.toml

johnthagen:
Refactor so that more configuration is in a single location

- https://docs.pytest.org/en/stable/reference/customize.html#pyproject-toml


dpgraham4401 (approved):
Thanks for the PR @johnthagen, seems like a no brainer.

auvipy (approved):

PR #872, opened 2025-05-15
Use generic types for get_objects_for_user

### LINKED ISSUE #864: Fix return type hint for `get_objects_for_user`
The return type hint for `get_objects_for_user` is currently marked as:

https://github.com/django-guardian/django-guardian/blob/6010336f1636afb4ace385f28edc39b40071319f/guardian/shortcuts.py#L439

But it really returns a `Queryset`, which can lead to false positive type issues.

Ideally, I think this should be a `Queryset[T]`, where the type is parameterized based on the input `klass`, but I'm not sure how tricky that would be to type annotate given how dynamic this function is.

- https://typing.python.org/en/latest/reference/generics.html

Ideally the following example would yield this return type:

```py
projects: QuerySet[Project] = get_objects_for_user(user, "view_project", Project.objects.all())
```

dpgraham4401:
## Changes

1. Adds a fix for `guradian.shortcuts.get_objects_for_user` type annotations, [see discussion in issue 846](https://github.com/django-guardian/django-guardian/issues/864).
2. Adds myself to the list of maintainers. 

closes #864 

Note: I initially also added tox to the `dev` dependency group but reverted that change.

### DIFF
--- a/guardian/shortcuts.py
+++ b/guardian/shortcuts.py
@@ -4,7 +4,7 @@ import warnings
 from collections import defaultdict
 from functools import partial, lru_cache
 from itertools import groupby
-from typing import Union, Any, Optional
+from typing import Union, Any, Optional, TypeVar, Type
 
 from django.apps import apps
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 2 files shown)


johnthagen (approved):

dpgraham4401:
There was an 500 HTTP status code when uv tried to install python 3.10 from [https://github.com/astral-sh/python-build-standalone](https://github.com/astral-sh/python-build-standalone/releases/download/20250409/cpython-3.10.17%2B20250409-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz). That's kind of annoying. I believe if the CI is run again the checks will pass.

johnthagen:
@thclark this would be a great MR to merge for a patch release.

thclark (approved):

PR #877, opened 2025-06-05
Dependency group separation

dpgraham4401:
Moves the database client dependency packages out of the `dev` group so the tox test suite can run without having mysql installed which is great for local development.


dpgraham4401:
@thclark I've reverted some of the formatting changes that my IDE auto added to make the diff easier to read. I would appreciate if we could merge this before #881 since it makes it difficult to work on guardian (on linux) while this is in main.

PR #878, opened 2025-06-12
Avoid a database hit

adrien-delhorme:
By checking `self.object` before calling `self.get_object` we can avoid a database hit.

### DIFF
--- a/guardian/mixins.py
+++ b/guardian/mixins.py
@@ -166,9 +166,10 @@ class PermissionRequiredMixin:
     def get_permission_object(self):
         if hasattr(self, 'permission_object'):
             return self.permission_object
-        return (hasattr(self, 'get_object') and self.get_object() or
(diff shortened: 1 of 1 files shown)


thclark:
Thanks for the contribution @adrien-delhorme. I don't see a linked issue for this PR, is there one?

I believe this is a breaking change because it alters the permission object that will be returned in some (many?) circumstances. The purpose of this mixin is to allow override of object selection by supplying a `get_object` method in the class this mixes into. So, if there's a database hit in your `get_object` method, the way to solve the hit is to either remove your `get_object` method (which will revert to the behaviour of returning `self.object`) or otherwise override it.

This PR would subvert the point of having the `get_object` method in the case where there's already an object attached.

I've labelled this tentatively as wont-fix for now, but if this hasn't solved your problem, please do raise an issue!

PR #888, opened 2025-07-01
Clean orphans with batching and time limited

Natgho:
While using it in a live environment, I noticed that deleting orphans took a long time and that we couldn't find where we left off when there were interruptions. 

Therefore, by imposing a time constraint, we can control when it will end.  

When the number of lines is very high, even with an iterator, since sorting is not possible and it cannot resume from where it left off, I integrated batch logic;
```python
>>> from utils.common import clean_orphan_obj_perms
>>> clean_orphan_obj_perms(max_duration_secs=3)
Time limit of 3s reached.
Finished orphan object permissions cleanup. Scanned: 414 | Removed: 0 | Batches processed: N/A
```
To prevent previously checked lines from being repeated, I added developer-friendly logs that inform the developer how to continue from where it left off.
```python
>>> clean_orphan_obj_perms(batch_size=1000, skip_batches=10, max_duration_secs=3)
Skipping batch 1 (size=1000)
Skipping batch 2 (size=1000)
Skipping batch 3 (size=1000)
Skipping batch 4 (size=1000)
...
Time limit of 3s reached.
Finished orphan object permissions cleanup. Scanned: 11000 | Removed: 448 | Batches processed: 1
To resume cleanup, call:
clean_orphan_obj_perms(batch_size=1000, skip_batches=11, max_duration_secs=3)
```
It could be further developed, but this is sufficient for now. We can say that one more “TODO” has been resolved.

### DIFF
--- a/guardian/testapp/tests/test_orphans.py
+++ b/guardian/testapp/tests/test_orphans.py
@@ -95,3 +95,66 @@ class OrphanedObjectPermissionsTest(TestCase):
             target.save()
             for perm in perms:
                 self.assertFalse(self.user.has_perm(perm, target))
+    
+    def test_clean_all_permissions(self):
+        """Should clean all orphaned permissions without batching."""
+        count = self._assign_and_delete_targets(3)
+        removed = clean_orphan_obj_perms()
+        self.assertEqual(removed, count)
+
+    def test_clean_with_batch_size(self):
+        """Should clean permissions in small batches."""
+        count = self._assign_and_delete_targets(5)
+        removed = clean_orphan_obj_perms(batch_size=2)
+        self.assertEqual(removed, count)
+
+    def test_clean_with_max_batches(self):
+        """Should stop after reaching max_batches."""
+        self._assign_and_delete_targets(5)
+        removed = clean_orphan_obj_perms(batch_size=1, max_batches=2)
+        self.assertEqual(removed, 2)
+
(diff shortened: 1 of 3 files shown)


Natgho:
@thclark Please pay attention here :)

thclark:
@Natgho thanks for the contribution. Is there an issue for this?

Natgho:
@thclark Your welcome :) 

No, we tried it with 18 million orphans row and it worked fine. and reduced the amount of data to 4.5 million. This resulted in significant DB efficiency.

My goal is to develop this function to enable multiprocessing after this PR is approved.

I think many guardian users are unaware of this function and the DB efficiency it provides.

thclark:
Tests are missing a method:

```
/Users/thc29/source/django-guardian/django-guardian/guardian/testapp/tests/test_orphans.py::OrphanedObjectPermissionsTest::test_clean_perms_command failed: self = <guardian.testapp.tests.test_orphans.OrphanedObjectPermissionsTest testMethod=test_clean_perms_command>

    def test_clean_perms_command(self):
        """Management command interface should also work."""
>       self._assign_and_delete_targets(2)
E       AttributeError: 'OrphanedObjectPermissionsTest' object has no attribute '_assign_and_delete_targets'

guardian/testapp/tests/test_orphans.py:153: AttributeError
```

Natgho:
I am going to create better version and the tests.

PR #889, opened 2025-07-08
default user model indexes are improved

Natgho:
In large databases, I noticed that queries were inefficient because these indexes were missing. These indexes ensure that constantly sent requests run quickly. (I use django-guardian together with Django-rest.)


Natgho:
Hi @thclark If you have time, could you take a look at the pending PRs? Implementing these performance improvements is of great importance for large systems.

PR #891, opened 2025-07-25
Investigating CI failures

thclark:


Natgho:
Checked, the problem is not related to the CI, discussed w @thclark

PR #902, opened 2025-09-02
Fix type and deprecate old name

Natgho:


Natgho:
There were conflicts here and backward compatibility had not been added. I fixed it with a new PR:
https://github.com/django-guardian/django-guardian/pull/902

Issue #458, opened 2016-09-05
cache get_anonymous_user()


pgeez:
`get_anonymous_user` does the same DB query every time the method called, though the query it executes is based on properties which are configured at application startup (namely `User.USERNAME_FIELD` and `guardian_settings.ANONYMOUS_USER_NAME`)

https://github.com/django-guardian/django-guardian/blob/devel/guardian/utils.py#L39

It seems like it should be possible to lazy-load and cache the anonymous user to avoid unneeded and repetitive lookups.

Thoughts?

ad-m:
@pgeez , good point. Do you can prepare pull requests to fix that?

pgeez:
Will do. Thanks.

Natgho:
https://github.com/django-guardian/django-guardian/pull/907
it is going to be add in soon. Sorry for keeping you waiting for 9 years :)

PR #907, opened 2025-09-08
Optional cached anonymous user

Natgho:
parametric anon user cache is developed (related issue: https://github.com/django-guardian/django-guardian/issues/458)

### DIFF
--- a/guardian/utils.py
+++ b/guardian/utils.py
@@ -5,6 +5,7 @@ Functions defined within this module are a part of django-guardian’s internal 
 and be considered unstable; their APIs may change in any future releases.
 """
 
+from functools import lru_cache
 import gc
 from itertools import chain
 import logging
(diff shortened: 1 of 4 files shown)


guardian/utils.py
```diff
@@ -5,6 +5,7 @@
 and be considered unstable; their APIs may change in any future releases.
 """
 
+from functools import lru_cache
```
thclark, line 8: As we learned from the issue we had with the LRU cache from the fixes we released in 3.1, we should always be using django's cache utilities for things like this. That way it remains a pluggable library and the cache performance isn't dependent on the app's deployment configuration.

Natgho: A fair point, that slipped my mind. Despite being a real pain in the code, especially in k8s environments, this damn thing is very useful :D

Natgho:
Lets continue here;
https://github.com/django-guardian/django-guardian/pull/912/files

PR #912, opened 2025-09-09
Optional cached anonymous user  - Pr 907

Natgho:


docs/configuration.md
```diff
@@ -154,6 +154,74 @@ queries or when latency is particularly important.
...
+```python
+# Cache anonymous user for 10 minutes
+GUARDIAN_CACHE_ANONYMOUS_USER = True
+GUARDIAN_ANON_CACHE_TTL = 600
+
+# Cache anonymous user for 1 hour
+GUARDIAN_ANON_CACHE_TTL = 3600
+
+# Cache anonymous user indefinitely (not recommended)
+GUARDIAN_ANON_CACHE_TTL = None
```
thclark: 1) If indefinite caching is not recommended, why would we allow it (is there a clear use case for it?)

2) There are two interacting settings where using one would be clearer. How about:
```py
# No cache (default)
GUARDIAN_ANONYMOUS_USER_CACHE_TTL = None

# Cache anonymous user for 1 hour
GUARDIAN_ANONYMOUS_USER_CACHE_TTL = 600

# Cache anonymous user indefinitely (not recommended)
GUARDIAN_ANONYMOUS_USER_CACHE_TTL = -1  
```
 (assuming we keep the indefinite possibility of course! The indefinite value could be `0` if you're old-school, but I don't recommend that in python if None is a default because boolean(None) == boolean(0))

docs/configuration.md
```diff
@@ -154,6 +154,74 @@ queries or when latency is particularly important.
...
+```
+
+!!! tip "Performance tuning"
+
+    - For high-traffic applications: Use shorter TTL (300-600 seconds)
+    - For stable applications: Use longer TTL (1800-3600 seconds)
+    - Consider your cache backend's memory limits when setting TTL
```
thclark, line 184: I like this advice, it always helps to have a bit of experience showing up in docs like this. Educate me @Natgho - why would you use a shorter TTL for high traffic apps?

docs/configuration.md
```diff
@@ -154,6 +154,74 @@ queries or when latency is particularly important.
...
+
+!!! note "Cache backend dependency"
+
+    This setting only takes effect when `GUARDIAN_CACHE_ANONYMOUS_USER`
+    is `True`. The actual caching behavior depends on your Django cache
+    configuration (`CACHES` setting).
```
thclark: You wouldn't have to have this if adopting the single setting approach

docs/configuration.md
Natgho: 1. The reason I am in favor of allowing unlimited cache here is that this decision should belong to the user, not us. Since the idea of preventing them from continuing this caching process permanently by managing it correctly is not appealing, I am in favor of granting such freedom.
2. Yes, I like that idea, so I'll do it :) “What's old is new again.” I will set the 0 status to be the same as the current None. This way, if a user activates the cache and specifies 0, we will accept unlimited cache.

docs/configuration.md
Natgho: My goal here is to reduce the number of requests to the database, despite the speed of SQL queries that handle a large volume of data in a short time. In other words, I'm working from the principle that “if traffic is very high, there will be more updates. save the select requests as possible as more.”
Example:
Consider a service that receives 100,000 requests per hour. Naturally, there will be frequent permission updates within these 100,000 requests. If we set the TTL too long, it would lead to unnecessary accumulation and extra cache updates. Therefore, in large systems, short-term caching makes more sense as a cumulative benefit.

docs/configuration.md
Natgho: In light of our previous discussions, this part will change.

thclark (approved):

docs/configuration.md
thclark: Thanks for the explanation @Natgho :)

Issue #770, opened 2022-01-30
create_anonymous_user() breaks on migration rollback


ZbigniewRA:
`post_migrate` handler `create_anonymous_user()` breaks when a migration that adds a field to the User model is rolled back.

Bug is in this line:
```
User.objects.using(kwargs['using']).get(**lookup)
```
The problem is that `get()` tries to fetch all fields, but some of the fields no longer exist (as they were removed in a rollback).

A fix is to fetch the only field we know exists:
```
User.objects.using(kwargs['using']).filter(**lookup).only(User.USERNAME_FIELD).get()
```

thclark:
Happy to accept a PR for this with a test, should anyone wish to contribute!

PR #919, opened 2025-09-15
rollback compatible create anon user fix is developed

Natgho:
https://github.com/django-guardian/django-guardian/issues/770

### DIFF
--- a/guardian/management/__init__.py
+++ b/guardian/management/__init__.py
@@ -28,7 +28,8 @@ def create_anonymous_user(sender, **kwargs):
         return
     try:
         lookup = {User.USERNAME_FIELD: guardian_settings.ANONYMOUS_USER_NAME}
-        User.objects.using(kwargs["using"]).get(**lookup)
+        # fixing #770
+        User.objects.using(kwargs["using"]).filter(**lookup).only(User.USERNAME_FIELD).get()
     except (User.DoesNotExist, DatabaseError):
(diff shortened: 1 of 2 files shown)


guardian/management/__init__.py
```diff
@@ -29,10 +29,18 @@ def create_anonymous_user(sender, **kwargs):
...
+        try:
+            retrieve_anonymous_function = import_string(guardian_settings.GET_INIT_ANONYMOUS_USER)
+            user = retrieve_anonymous_function(User)
+            user.save(using=kwargs["using"])
+        except DatabaseError:
+            # If we still get a DatabaseError when trying to save,
+            # it means the table doesn't exist (rollback scenario)
+            # In this case, we should silently return as the migration
+            # will handle user creation when it's run again
+            return
```
thclark: This doesn't seem to address the issue in #770 - I was expecting to see line 31
```
User.objects.using(kwargs["using"]).get(**lookup)
```
become:
```
User.objects.using(kwargs['using']).filter(**lookup).only(User.USERNAME_FIELD).get()
```

This seems to solve a different thing - table doesn't exist errors. @Natgho have you encountered this DatabaseError in rollbacks in the wild?

Natgho: when looking at this error, I realized there is an another potential bug here and first fixed the missing table existing check, but forgot to change that line in the process. I've corrected it.

thclark: Great, thanks

thclark (approved):

PR #920, opened 2025-09-15
generator warnings and logics are added for security purposes

Natgho:


guardian/mixins.py
```diff
@@ -161,6 +162,12 @@ def get_required_permissions(self, request: Optional[HttpRequest] = None) -> lis
         """
         if isinstance(self.permission_required, str):
             perms = [self.permission_required]
+        elif isinstance(self.permission_required, GeneratorType):
+            raise ImproperlyConfigured(
+                "'PermissionRequiredMixin' does not support generators for "
+                "'permission_required' attribute as they can only be consumed once, "
+                "potentially leading to security issues. Use a list or tuple instead."
+            )
```
thclark: Raising exceptions will burn people using generators in production. Could we add a warning and a deprecation instead, then in v4 we can reinstate this error to prevent the insecure behaviour?

Natgho: make sense, okay, I will 👍

thclark: thanks!

thclark (approved):

Issue #644, opened 2019-09-27
Using inlines in Django Admin with guardian


elimayor:
Are inline fields not supported in Django Admin with Guardian? 

```
class PhoneInline(admin.StackedInline):
    model = UserPhone
    extra = 1

class UserAdmin(GuardedModelAdmin):
    inlines=[PhoneInline]
    list_display = ('Username', 'FirstName', 'LastName')
    search_fields = ['Username','FirstName','LastName']
```

However the inline forms no longer appear.

solace:
Bumping this as it's over a year old as I'm interested in this functionality as well.

solace:
Traced this to [django/contrib/admin/options.py](https://github.com/django/django/blob/fed8129276eac973fd48a2f33b7e1735a866628d/django/contrib/admin/options.py#L596).

You need to implement the following on the inline:
```
has_view_or_change_permission
has_add_permission
has_delete_permission
```

At least, that's what resolved it for me.

SimonF89:
any news here? Would be important for my current projecT!

SimonF89:
@solace could you add a snipped? I dont know how to acces the inline instance to get the True boolean if the permission is granted.

solace:
@SimonF89 Sorry about the delay, it's been a while since I've touched this codebase so I had to forage.

I have a mixin that I use with guarded inlines that has this structure. There is a possibility that some of these are custom perm checks, but I can't recall which ones, so bear that in mind.

```
class GuardedInlineAdminMixin:
    # ... 
    def has_access_group_permission(self, request):
        return # tests request for access_group permissions

    def has_write_group_permission(self, request):
        return # tests request for write_group permissions

    def has_own_permission(self, request, obj=None):
        return # tests request for own permissions

    def has_add_permission(self, request, obj=None):
        return # tests request for add permissions

    def has_view_permission(self, request, obj=None):
        return # tests request for view permissions

    def has_change_permission(self, request, obj=None):
        return # tests request for change permissions

    def has_delete_permission(self, request, obj=None):
        return # tests request for delete permissions

    def has_module_permission(self, request, obj=None):
        return # tests request for module permissions
```

I checked the Admin class though and it uses `GuardedModelAdminMixin` not `GuardedModelAdmin`, and the perms mixin has to go after it.

eg.

```
class CustomAdmin(guardian_admin.GuardedModelAdminMixin, GuardedInlineAdminMixin):
    ...
```

Hope that helps.

PR #921, opened 2025-09-15
644 using inlines in django admin with guardian

Natgho:
https://github.com/django-guardian/django-guardian/issues/644

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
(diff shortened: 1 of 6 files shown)


guardian/admin.py
```diff
@@ -23,6 +23,73 @@
...
+    def has_add_permission(self, request, obj=None):
+        """Check if the user has permission to add instances of this inline model."""
+        opts = self.model._meta
+        codename = f"add_{opts.model_name}"
+        perm = f"{opts.app_label}.{codename}"
+
+        if obj is None:
+            # For global permissions when obj is None
+            return request.user.has_perm(perm)
+        else:
```
thclark, line 49: @Natgho I have to leave for work but later am going to check if this is consistent with the rest of guardian functionality. There are a lot of issues surrounding whether globals should be included or excluded; we should work toward whatever's consisten

Natgho: Of course, there's no need to rush.

Actually, in another issue, I saw the danger of this authority problem. It's not on my to-do list yet, but this issue has nothing to do with that bug. I'll deal with that issue separately. There was a ticket with a screenshot in it; I'll share that too when I find it.

https://github.com/django-guardian/django-guardian/issues/756

thclark: OK, yes, checked and I think this is the right way of doing it; otherwise any global permissions will be wiped out.

thclark (approved):

PR #923, opened 2025-09-18
786 get perms returns empty list while get user perms returns permissions

Natgho:


guardian/shortcuts.py
```diff
@@ -227,34 +227,116 @@ def remove_perm(
...
+        - For inactive users (is_active=False), returns empty QuerySet
+        - Return type is QuerySet, not list (unlike get_perms())
+
+    Common Confusion:
+        Many users expect this function to return the same results as get_perms(),
+        but that's incorrect. Here's the difference:
+
+        - get_perms(user, obj): Returns ALL permissions (user + group permissions)
+        - get_user_perms(user, obj): Returns ONLY direct user permissions
+        - get_group_perms(user, obj): Returns ONLY group permissions
```
dpgraham4401: @Natgho it probably makes more sense to put this information in it's own section in documentation directly since it's relevant to multiple functions instead of putting it in the docstring of just one of the relevant functions.

dpgraham4401: Same goes for the new uses cases section and important notes. 

Nice addition 👍

guardian/core.py
```diff
@@ -123,6 +123,11 @@ def get_user_filters(self, obj: Model):
         return user_filters
 
     def get_user_perms(self, obj: Model) -> QuerySet[Permission]:
+        if self.user and not self.user.is_active:
+            from django.db.models import QuerySet
```
dpgraham4401: Why import inside the function? @Natgho move to top import block, thanks!

dpgraham4401: Actually, it looks like it's already imported, is there a reason to re-import?

guardian/core.py
```diff
@@ -123,6 +123,11 @@ def get_user_filters(self, obj: Model):
...
+
+            return Permission.objects.none().values_list("codename", flat=True)
```
dpgraham4401, line 127: This looks like it will return the same thing as original return statement. Why the added check? I'm not seeing anything in the PR description for the reason why

guardian/shortcuts.py
```diff
@@ -227,34 +227,116 @@ def remove_perm(
...
+        - get_user_perms(user, obj): Returns only direct user permissions
+        - get_group_perms(user, obj): Returns only group permissions
+
+        For any user: get_perms(user, obj) = get_user_perms(user, obj) + get_group_perms(user, obj)
+
+    Use Cases:
+        - When you need to see what permissions come from group membership
+        - When implementing permission management interfaces that show source of permissions
+        - When auditing group-based permissions
+        - When you need to understand permission inheritance
```
dpgraham4401: Same comment as on `get_perms` , let's move these to sections in the docs so it's all in one place and we're not repeating ourselves in all the docstrings.

guardian/core.py
```diff
@@ -123,6 +123,11 @@ def get_user_filters(self, obj: Model):
         return user_filters
 
     def get_user_perms(self, obj: Model) -> QuerySet[Permission]:
+        if self.user and not self.user.is_active:
```
dpgraham4401, line 126: What if you need to get the permissions of an inactive user?

guardian/core.py
Natgho: right, removed.

guardian/core.py
Natgho: There is a performance improvement here;
“If the user is inactive, instead of going to the database and checking, it directly returns an empty list using ‘none()’.”

guardian/core.py
Natgho: This is the default behavior within this DG.

Here is some other examples;
https://github.com/django-guardian/django-guardian/blob/a2b1b11901be5b665faf0d2561bdedf09ddb78f0/guardian/core.py#L76-L76
https://github.com/django-guardian/django-guardian/blob/a2b1b11901be5b665faf0d2561bdedf09ddb78f0/guardian/core.py#L164-L165

guardian/shortcuts.py
Natgho: moved.

guardian/shortcuts.py
Natgho: moved under the userguide part.

PR #924, opened 2025-09-18
group permissions check is added

Natgho:


Natgho:
> This looks good to me @Natgho, only one question.
> 
> should we implement `has_module_perms` since we're just checking if they're a superuser which already checked by Django's model backend?

Actually, this is the right approach. Django already provides this control. Offering it as the default seems a bit like imposing it. If the user wants, they can solve this with the default Django structure. I'm removing this section.

dpgraham4401:
One last thought, I think we might have to consider this a breaking change 😬. 

Since, Django call this method if it's available, existing Django views may have the additional permissions added from the new group check before Django checks if the user is authorized. 

It seems inline with what guardian should be doing, but since it wasn't doing it before, it may be wise to mark as a breaking change.

Natgho:
> One last thought, I think we might have to consider this a breaking change 😬.
> 
> Since, Django call this method if it's available, existing Django views may have the additional permissions added from the new group check before Django checks if the user is authorized.
> 
> It seems inline with what guardian should be doing, but since it wasn't doing it before, it may be wise to mark as a breaking change.

@thclark please don't forget to this info when we are going to release :)

PR #926, opened 2025-09-18
Code coverage system is added

Natgho:


Natgho:
> Looks good from our perspective, thank you very much!

Hi, thank you for your review. Well then, let's add it to the system! (=

Issue #532, opened 2017-11-15
Document how to apply permissions "as a template"


elcolie:
Request new feature
**Requirements:**
[ ] Be able to transfer `permissions` to new `group` by API

ad-m:
What do you mean by "transfer" of permission? There is API to (un)assign permissions.

elcolie:
**Background:**
1. I have multiple companies in the system
1. One company has `Admin, Manager, Front, Support`. Let's call them `Role`
1. One `Role` has One `Django Group`. (I define)
1. One groups has multiple permissions. For example (P1, P2, ... , Pn)

**Situation**
Role `Admin` has (P1, P2, P3)
1. Add new `Admin2` Role with permission (P1, P2)
1. Human customer remove role `Admin`. 
1. All the instances under the `Admin` are transfered to `Admin2`

If `django-guardian` has simple API to do transfer would be nice.
Right now what come up in my idea is 
1. List all the instances
1. Assign them to the new `Group` one by one

Allan-Nava:
I need the same feature for a "group type" such as staff group for a group. Only the staff member have permission for the group

thclark:
I don't think there's a clear use case for transferring perms from one object to another, however I understand @Allan-Nava 's point that a "template" of permissions would be very helpful (and actually have implemented something quite similar in one of my recent projects).

I think an example project showing how such related groups could have a template of permissions set is a good plan for the documentation. In terms of actually implementing it, #779 is very helpful because it'll facilitate bulk creation.

Natgho:
I liked this idea and created a PR implementing it, refining the concept a bit further by including copy and transfer features not only at the group level but also at the user level.
https://github.com/django-guardian/django-guardian/pull/927

dpgraham4401:
This strikes me as a relatively niche feature; something that few applications will need and the ones that do will likely each need it done in a different ways.

I think we could document this (like the issue title suggests) and provide an example without needing to maintain something that few people will need.

Natgho:
> This strikes me as a relatively niche feature; something that few applications will need and the ones that do will likely each need it done in a different ways.
> 
> I think we could document this (like the issue title suggests) and provide an example without needing to maintain something that few people will need.

Actually, what is wanted is not something extremely complex. Of course, there are limitations. If the idea of “let's just document it” prevails, we'll take the documentation part and move on.

[...]

PR #927, opened 2025-09-18
532 document how to apply permissions as a template

Natgho:


guardian/shortcuts.py
dpgraham4401, line 1: @Natgho I don't believe a consensus was reached in #532  that this was a feature we wanted to support. We can add something to the documentation (like the title of your PR suggests) without expanding guardian's API, but this 110+ long behemouth of a function is going to have to be maintained.

Natgho: I am aware that this process can be dangerous in some cases, so I create quite a variety of tests. 

But the critical thing here is that this feature does not harm DG's default structure or backbone. It is added as an additional feature that users can use whenever they want. 

So, if they encounter an error in scenarios we cannot foresee at the moment, they can open an issue, and we can make it perfect (if it's not perfect yet) over time, right?

dpgraham4401: >But the critical thing here is that this feature does not harm DG's default structure or backbone. It is added as an additional feature that users can use whenever they want.

Creating something is easy, but as you've seen from #932, ensuring there are people who can maintain it when something goes wrong is not a given. When we publish new features, it's not just a "additional", peripheral-style feature, it becomes part of guardian public API and we implicitly create a contract that we're agreeing to maintain this. 

> So, if they encounter an error in scenarios we cannot foresee at the moment, they can open an issue, and we can make it perfect (if it's not perfect yet) over time, right?

Right now guardian **_works_**, and since it's a package that provides the authorization for many pubic facing applications (including ones I work with), it's critical that this fact doesn't change. "encountering an error" as you say, means creating security risks, and a boat load of headache or worse.

> and we can make it perfect (if it's not perfect yet) over time, right?

Do you think that writing (or have AI write for you) 200+ line functions is going to be easy to understand and change in the future? Also, when we fix things, we may be changing a behavior that other rely on.

guardian/testapp/tests/test_shortcuts.py
```diff
@@ -1377,3 +1381,549 @@ class ContentTypeCacheTestCase(ContentTypeCacheMixin, TestCase):
 
 class ContentTypeCacheTransactionTestCase(ContentTypeCacheMixin, TransactionTestCase):
     """Test cache against TransactionTestCase"""
+
+
+class TransferGroupPermsTest(TestCase):
```
dpgraham4401, line 1386: I'm only seeing test that deal with permissions defined via django's `contenttype` framework, what about custom groups and permission models?

guardian/shortcuts.py
Natgho: > Creating something is easy, but as you've seen from https://github.com/django-guardian/django-guardian/discussions/932, ensuring there are people who can maintain it when something goes wrong is not a given. When we publish new features, it's not just a "additional", peripheral-style feature, it becomes part of guardian public API and we implicitly create a contract that we're agreeing to maintain this.

I certainly understand your concern about maintenance. However, aren't all open-source projects developed with the assumption that they will be maintained in the future? Otherwise, since the project consists of thousands of lines of code, shouldn't we just leave it alone?

The passivity regarding maintenance up until now is a wound for DG. I accept that. However, I am trying to be here, and I will try to be here unless something extra comes up. Of course, I will respond quickly at times and slowly at others, just like any other mortal. Still, I can promise that it will not turn into a repository that has not been touched for years, as it was in the past, as long as I am alive.
> Right now guardian works, and since it's a package that provides the authorization for many pubic facing applications (including ones I work with), it's critical that this fact doesn't change. "encountering an error" as you say, means creating security risks, and a boat load of headache or worse.

When I say this feature is not a flow-breaker, my intention is to point out that it does not interfere with existing permission assignments/permission arrangements/permission updates, but only provides a new gain. Users can continue to make assignments as they do now instead of using this function.

>Do you think that writing (or have AI write for you) 200+ line functions is going to be easy to understand and change in the future? Also, when we fix things, we may be changing a behavior that other rely on. 

You're already reviewing the code, yes, the line count is high, but that doesn't mean the code is complex or unreadable, does it? Someone competent in this area, upon seeing this code, can easily maintain it. We write it adhering to principles like SOLID/DRY/KISS etc. and by testing it. Our current code coverage is 98% (please take a look at that PR and let's make it the default for the DG :) ). Those who want to contribute to the project can easily understand many things just by reading the tests.

guardian/testapp/tests/test_shortcuts.py
Natgho: For these functions, we should assume that the developer is using the default group and permission models. I can add a warning for this. what do you think?

Natgho: Default permission check is added.

Natgho:
Hi @dpgraham4401 , what do you think about the final version? Do you still think this feature is too risky? I think it's stable enough and seems like a solid feature for the next version. But if you still think it's too risky, I won't merge this PR if Tom sees it and agrees with you.

dpgraham4401:
> what do you think about the final version? Do you still think this feature is too risky? I think it's stable enough and seems like a solid feature for the next version. But if you still think it's too risky, I won't merge this PR if Tom sees it and agrees with you.

Here's some of the reason's why I don't think we should move forward with this:

1. It doesn't work with the other supported features (custom permission models, something that's used in all projects I've worked with that use guardian).

2. It's niche enough of a feature (made further niche by point 1) that many will likely never need/want it. [The issue has been open for ~9 years and there's been no additional users supporting the idea, emoji reactions, are anything](https://github.com/django-guardian/django-guardian/issues/532). As an open source package, we should be implementing common **need**s, not something that only 1 or 2 **want**. The issue only ask that we document an example as well. 

    - Those that do need this will likely need a variant of it e.g., copy only read-only permissions, don't copy permissions for inactive users, etc. etc. These are just example off the cuff, but the point is things tend to get more complicated and we have to support these once we commit to it.

3. Even if it's stable today, it's will require maintenance tomorrow. Whoever is maintaining at the time will need to support this feature for the select few who need this. 

It's OK to say "no." We don't need to support every feature request ever submitted. I would argue it's the responsible thing to do in this case. I don't think the juice is worth the squeeze.

Natgho:
> > what do you think about the final version? Do you still think this feature is too risky? I think it's stable enough and seems like a solid feature for the next version. But if you still think it's too risky, I won't merge this PR if Tom sees it and agrees with you.
> 
> Here's some of the reason's why I don't think we should move forward with this:
> 
> 1. It doesn't work with the other supported features (custom permission models, something that's used in all projects I've worked with that use guardian).
> 2. It's niche enough of a feature (made further niche by point 1) that many will likely never need/want it. [The issue has been open for ~9 years and there's been no additional users supporting the idea, emoji reactions, are anything](https://github.com/django-guardian/django-guardian/issues/532). As an open source package, we should be implementing common **need**s, not something that only 1 or 2 **want**. The issue only ask that we document an example as well.
>    
>    * Those that do need this will likely need a variant of it e.g., copy only read-only permissions, don't copy permissions for inactive users, etc. etc. These are just example off the cuff, but the point is things tend to get more complicated and we have to support these once we commit to it.
> 3. Even if it's stable today, it's will require maintenance tomorrow. Whoever is maintaining at the time will need to support this feature for the select few who need this.
> 
> It's OK to say "no." We don't need to support every feature request ever submitted. I would argue it's the responsible thing to do in this case. I don't think the juice is worth the squeeze.

Okay, based on your level of commitment and experience in this matter, I'm closing this feature I developed. @thclark, you can check it afterwards if you want.

thclark:
Sorry, I've utterly pretty snowed under with work at the moment but I agree with @dpgraham4401 in this case; it's too big and too niche a feature. Nevertheless @Natgho that doesn't discount the value of you forging on with it, well done and thank you - such a great spirit :)

dpgraham4401:
> Nevertheless @Natgho that doesn't discount the value of you forging on with it, well done and thank you - such a great spirit :)

I second this! Thank you @Natgho for putting in the time

PR #939, opened 2025-10-18
Fix small typo

jeverling:


Natgho:
Thanks for fixing the typo 🥇

Natgho:
image
Dockerhub is down for now. CI Pipeline is crashing because of this. I will check it out when the DH is available.

Issue #938, opened 2025-10-17
PGSQL can't compare UUID and char


jucajuca:
**Describe the bug**

When using an ULID, it seems that guardian can't create a valid query. 

Passing a query also had the same error. I am not sure if I am missing something.

Example model:

```python
from django_ulid.models import ULIDField
from django_ulid.models import default as default_ulid

def new_ulid():
    return str(default_ulid())

class MyModel(models.Model):
    id = ULIDField(primary_key=True, default=new_ulid, editable=False)

```

Logs from SQL
```
2025-10-17 20:03:37.378 UTC [211] ERROR:  operator does not exist: character varying = uuid at character 454
2025-10-17 20:03:37.378 UTC [211] HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
2025-10-17 20:03:37.378 UTC [211] STATEMENT:  SELECT "my_model"."id", "my_model"."slug_id", "my_model"."is_public", "my_model"."linked_data", "my_model"."created_at", "my_model"."updated_at" FROM "my_model" WHERE ("my_model"."id" IN (SELECT V0."object_pk" FROM "guardian_userobjectpermission" V0 INNER JOIN "auth_permission" V2 ON (V0."permission_id" = V2."id") WHERE (V0."user_id" = 1 AND V2."content_type_id" = 17 AND V0."object_pk" IN (SELECT U0."id" FROM "my_model" U0) AND V2."codename" IN ('view_catalogue'))) OR "my_model"."id" IN (SELECT V0."object_pk" FROM "guardian_groupobjectpermission" V0 INNER JOIN "auth_permission" V2 ON (V0."permission_id" = V2."id") WHERE (V0."group_id" IN (SELECT U0."id" FROM "auth_group" U0 INNER JOIN "users_user_groups" U1 ON (U0."id" = U1."group_id") WHERE U1."user_id" = 1) AND V2."codename" IN ('view_catalogue') AND V2."content_type_id" = 17 AND V0."object_pk" IN (SELECT U0."id" FROM "my_model" U0)))) LIMIT 1

```

DB:

<img width="609" height="600" alt="Image" src="https://github.com/user-attachments/assets/b3e8f3d8-fd0e-47d1-a507-eeab65627d68" />

**To Reproduce**
1. Create a model with an ULID
2. Try using get_objects_for_user

**Desktop (please complete the following information):**
django==5.0.4  # pyup: < 5.0           # https://www.djangoproject.com/
[...]

PR #940, opened 2025-10-20
UlidField support is added

Natgho:


guardian/shortcuts.py
```diff
@@ -889,6 +889,12 @@ def _handle_pk_field(queryset):
             output_field=CharField(),
         )
 
+    # Handle ULID fields and other CharField-based primary keys
+    # ULID fields are typically CharField-based and should be cast to CharField for consistent comparison
+    # https://github.com/django-guardian/django-guardian/issues/938
+    if isinstance(pk, CharField) or (hasattr(pk, "__class__") and "ULID" in pk.__class__.__name__):
+        return partial(Cast, output_field=CharField())
```
thclark, line 896: This is a breaking change because anybody already using any field with ULID in the class name now casts their pk instead of returning None as expected. I'd really rather support only django core here. If we were to support other pk field types, it should be via a custom resolver (plugin-style)

guardian/testapp/models.py
```diff
@@ -128,6 +128,34 @@ class UUIDPKModel(models.Model):
...
+    return str(uuid.uuid4()).replace("-", "")[:26].upper()
+
+
+class ULIDPKModel(models.Model):
+    """
+    Model for testing whether get_objects_for_user will work when the objects to
+    be returned have ULID primary keys (CharField-based ULIDs).
+    """
+
+    ulid_pk = ULIDField(
```
thclark, line 152: ULIDFIeld is not a dev dependency (it would have to be for this to work - but I don't think it should be, see comments)

dpgraham4401:
> Buuuut…. ULIDField is from django-ulid - a pretty rare library. The fact that this hasn't shown up in a decade-plus suggests this is a real edge case. It seems maintained right now but if the library got deprecated we'd have to make another breaking change to guardian to remove it.

I agree with Tom. Based on the code snippet, It looks like the author of #938 actually using [`django-ulid`](https://github.com/ahawker/django-ulid) (which is not maintained, and not widely adopted) but they listed [`python-ulid-django`](https://github.com/Xdynix/python-ulid-django) (maintained, but like Tom said, not used).

There should be a way to override a guardian method to use [Cast](https://docs.djangoproject.com/en/5.2/ref/models/database-functions/#cast) but without knowing more about how the author's trying to use guardian, it's tough to know where that would need to happen. But again, something that would fall under the application specific, not inside the scope of guardian.

Natgho:
Actually, my goal in making this improvement was simply to support a slightly more advanced version of CharField. That's why I didn't hesitate to add it. However, considering the “less is more!” principle, I agree with what you said. Perhaps in the future, the library will be used more and have a maintainer. Then we will provide official support. For now, I'm marking the PR as unresolved and closing it.

On the other hand, while @tom's ideas are nice, as @dpgraham said, these solutions are overly flexible. If you start with the default settings, for example, and then want to change “_handle_pk_field” later, a project can easily break. Therefore, developing this idea will be quite difficult and complex. Maybe we can discuss it after refining this idea. In its current state, developing it would not be functional.

I'm marking the PR as “will not be done” and closing it.

PR #941, opened 2025-10-21
multilanguage support is added

Natgho:


Natgho:
Fixing is harder than rewriting, I'll rewrite this :)

PR #943, opened 2025-10-22
Add adopters section

### LINKED ISSUE #679: Adopters
Django-guardian is used in very many projects (GitHub suggests that there are over 2,000) - both commercial and non-commercial, both closed-source and open-source.

I think it will be helpful to show these projects in the README library / separate documentation section to inspire the confidence of potential library users regarding its adaptation in the Django community and its high maturity.

This can contribute to library marketing, but also - in the case of open-source projects - to inspire users who are trying to understand how to use it. Maybe this will help when designing the library API in the future.

To start with, I suggest in the comments to this issue to share the names of the projects and, if possible, links to them. After some time it will be collected and published in the right place.

Natgho:


docs/who-uses-guardian.md
```diff
@@ -0,0 +1,143 @@
...
+Feel free to reach out through our community channels.
+
+## Implementation Patterns
+
+Based on real-world usage, here are common implementation patterns:
+
+### Multi-Tenant Applications
+```python
+# Tenant-isolated permissions
+assign_perm('view_document', user, document, tenant_scope=request.tenant)
```
dpgraham4401: So we're just making up stuff in the documentation now?

dpgraham4401 (changes requested):

dpgraham4401:
How verbose does an "adopters" section need to be? Why are we using LLMs to dilute our documentation with low value content?

When was the last time you were at work and needing documentation for a package and thought to yourself "_well I came here to read about this class's attributes but this LLM generated list of random industries is way cool, I'm gonna read that instead!_"?


On another note, who in their right mind would advertise that they're using guardian for security? That's like advertising that you're vulnerable next time there's [another security CVE from Django guardian](https://security.snyk.io/vuln?search=Django-Guardian)?

Natgho:
> How verbose does an "adopters" section need to be? Why are we using LLMs to dilute our documentation with low value content?

There is no standard in this regard. I have removed the unnecessary sections. Open to improvement.
 
> When was the last time you were at work and needing documentation for a package and thought to yourself "_well I came here to read about this class's attributes but this LLM generated list of random industries is way cool, I'm gonna read that instead!_"?

Maybe you should calm down a bit?
I created the content of the document as a starting point. this section should exist not for integration, but rather as a section for companies to add themselves if they wish. I cleaned up the unnecessary/incorrect parts. If you can do better, please don't hesitate to update or suggest improvements.

> On another note, who in their right mind would advertise that they're using guardian for security? That's like advertising that you're vulnerable next time there's [another security CVE from Django guardian](https://security.snyk.io/vuln?search=Django-Guardian)?

People may want to mention that they use this project because it is “open source.” Or companies may want to specify which projects they use, or simply write their names here by saying “we use” in general. Companies that think like you say, that have concerns, won't add their names here, and that's that?

thclark (approved):
I'm OK with this, I think it'd be good to whip up a list of a few of the example cases, rather than just giving users a template, but that can follow on (I'll add my own next time I'm doing it)... but I don't think anything in there is an exaggeration at all.

dpgraham4401:
@thclark the conversation comes from the original proposed changes which included completely AI-hallucinated guardian functionality.

PR #946, opened 2025-11-10
python 3.14 support is added

Natgho:


.github/workflows/coverage.yml
```diff
@@ -46,12 +46,33 @@ jobs:
...
+        if command -v python3.14 &> /dev/null && python3.14 --version; then
+          echo "version=3.14" >> $GITHUB_OUTPUT
+          echo "detected_python=3.14" >> $GITHUB_OUTPUT
+          echo "Python 3.14 is available"
+        else
+          echo "version=3.13" >> $GITHUB_OUTPUT
+          echo "detected_python=3.13" >> $GITHUB_OUTPUT
+          echo "Python 3.14 not available, falling back to 3.13"
+        fi
+
```
thclark, line 73: I'd suggest keeping it simple here. If the installation of 3.14 fails (intermittently or otherwise), we need to find out why. Branching the test runner to account for it instead makes things very complex to debug on failure... Have you been experiencing times when 3.14 installation failed on this step?

.github/workflows/coverage.yml
```diff
@@ -61,9 +82,21 @@ jobs:
...
+      env:
+        TOX_GH_MAJOR_MINOR: '3.14'
+        PYTEST_ADDOPTS: -vv --durations=20
+        DIFF_AGAINST: HEAD
+        UV_PYTHON_PREFERENCE: only-managed
+      run: tox run -vv --skip-pkg-install
 
-    - name: Run tests with coverage
+    - name: Run tests with coverage (Python 3.13 fallback)
+      if: steps.python_version.outputs.version == '3.13' || steps.coverage_314.outcome == 'failure'
```
thclark, line 99: See my above comment, because of bypassing the actual root problem it makes the workflow more complicated.

thclark (changes requested):
Thanks for the contribution @Natgho - however, branching makes the workflow significantly more complicated. Could we simply shift to 3.14? It feels to me like if there's an intermittent error installing it, we should be attacking that rather than making the workflow more complex to account for it?

Natgho:
Due to compatibility issues, dependencies also need to start 3.14 support, so it has been temporarily closed.

PR #948, opened 2025-11-12
Inherit ObjectPermissionBackend from Django's BaseBackend

### LINKED ISSUE #947: Make ObjectPermissionBackend inherit from BaseBackend
**Is your feature request related to a problem? Please describe.**
I'm trying to use `django-guardian` in an `async` context, but calling `await user.ahas_perm` won't work since `ObjectPermissionBackend` does not implement any of the `async` methods.

**Describe the solution you'd like**
[`django.contrib.auth.backends.BaseBackend`](https://docs.djangoproject.com/en/5.2/topics/auth/customizing/#writing-an-authentication-backend) is a base class for authentication backends. It provides a default `async` versions of every required method by wrapping them in a `sync_to_async` call.

Inheriting `ObjectPermissionBackend` from `BaseBackend` would better document that it is in fact an authentication backend, and would provide a crude version of those `async` methods.

**Describe alternatives you've considered**
Implementing those `async` methods (without using `sync_to_async`) would be the ideal, but it's a long way. Starting by using these default async versions is how Django itself has been slowing achieving async support.

**Additional context**

hartungstenio:
#947

### DIFF
--- a/guardian/backends.py
+++ b/guardian/backends.py
@@ -1,6 +1,6 @@
 from typing import Any, Iterable, Optional
 
-import django
+from asgiref.sync import sync_to_async
 from django.contrib.auth import get_user_model
 from django.contrib.auth.backends import BaseBackend
 from django.db import models
@@ -111,38 +111,35 @@ class ObjectPermissionBackend(BaseBackend):
         check = ObjectPermissionChecker(user_obj)
         return check.has_perm(perm, obj)
 
-    if django.VERSION >= (5, 2):
+    async def ahas_perm(self, user_obj: Any, perm: str, obj: Optional[Model] = None) -> bool:
+        """Check if a user has the permission for a given object.
 
-        async def ahas_perm(self, user_obj: Any, perm: str, obj: Optional[Model] = None) -> bool:
-            """Check if a user has the permission for a given object.
+        Async version of :meth:`ObjectPermissionBackend.has_perm`.
 
-            Async version of :meth:`ObjectPermissionBackend.has_perm`.
+        Returns `True` if given `user_obj` has `perm` for `obj`.
+        If no `obj` is given, `False` is returned.
+        The main difference between Django's `ModelBackend` is that we can pass
+        `obj` instance here and `perm` doesn't have to contain
+        `app_label` as it can be retrieved from given `obj`.
 
-            Returns `True` if given `user_obj` has `perm` for `obj`.
-            If no `obj` is given, `False` is returned.
-            The main difference between Django's `ModelBackend` is that we can pass
-            `obj` instance here and `perm` doesn't have to contain
-            `app_label` as it can be retrieved from given `obj`.
+        **Inactive user support**
 
-            **Inactive user support**
+        If `user` is authenticated but inactive at the same time, all checks
+        always return `False`.
 
-            If `user` is authenticated but inactive at the same time, all checks
-            always return `False`.
+        Note:
+            Remember, that if user is not *active*, all checks would return `False`.
 
-            Note:
-               Remember, that if user is not *active*, all checks would return `False`.
+        Parameters:
+            user_obj (User): User instance.
+            perm (str): Permission string.
+            obj (Model): Model instance.
 
-            Parameters:
-                user_obj (User): User instance.
-                perm (str): Permission string.
-                obj (Model): Model instance.
+        Returns:
+            `True` if `user_obj` has permission, `False` otherwise.
+        """
 
-            Returns:
-                `True` if `user_obj` has permission, `False` otherwise.
-            """
-            from asgiref.sync import sync_to_async
-
-            return await sync_to_async(self.has_perm)(user_obj, perm, obj)
+        return await sync_to_async(self.has_perm)(user_obj, perm, obj)
 
     def get_user_permissions(self, user_obj: Any, obj: Optional[Model] = None) -> Iterable[str]:
         """Returns user permissions for a given object.
@@ -155,11 +152,6 @@ class ObjectPermissionBackend(BaseBackend):
         Returns:
              a set of permission strings that the given specific `user_obj` has for `obj`.
         """
-        # This backend only handles object-level permissions
-        if obj is None:
-            return set()
-
-        # check if user_obj and object are supported
         support, user_obj = check_support(user_obj, obj)
         if not support:
             return set()
--- a/guardian/backends.py
+++ b/guardian/backends.py
@@ -1,9 +1,9 @@
 from typing import Any, Iterable, Optional
 
 from django.contrib.auth import get_user_model
+from django.contrib.auth.backends import BaseBackend
 from django.db import models
 from django.db.models import Model
-from django.http import HttpRequest
 
 from guardian.conf import settings
 from guardian.core import ObjectPermissionChecker
@@ -55,15 +55,12 @@ def check_support(user_obj: Any, obj: Model) -> Any:
     return obj_support and user_support, user_obj
 
 
-class ObjectPermissionBackend:
+class ObjectPermissionBackend(BaseBackend):
     """Django backend for checking object-level permissions."""
 
     supports_object_permissions = True
     supports_anonymous_user = True
     supports_inactive_user = True
-
-    def authenticate(self, request: HttpRequest, username: Optional[str] = None, password: Optional[str] = None) -> Any:
-        return None
 
     def has_perm(self, user_obj: Any, perm: str, obj: Optional[Model] = None) -> bool:
         """Check if a user has the permission for a given object.
--- a/guardian/backends.py
+++ b/guardian/backends.py
@@ -1,5 +1,6 @@
 from typing import Any, Iterable, Optional
 
+import django
 from django.contrib.auth import get_user_model
 from django.contrib.auth.backends import BaseBackend
 from django.db import models
@@ -110,6 +111,39 @@ class ObjectPermissionBackend(BaseBackend):
         check = ObjectPermissionChecker(user_obj)
         return check.has_perm(perm, obj)
 
+    if django.VERSION >= (5, 2):
+
+        async def ahas_perm(self, user_obj: Any, perm: str, obj: Optional[Model] = None) -> bool:
+            """Check if a user has the permission for a given object.
+
+            Async version of :meth:`ObjectPermissionBackend.has_perm`.
+
+            Returns `True` if given `user_obj` has `perm` for `obj`.
+            If no `obj` is given, `False` is returned.
+            The main difference between Django's `ModelBackend` is that we can pass
+            `obj` instance here and `perm` doesn't have to contain
+            `app_label` as it can be retrieved from given `obj`.
+
+            **Inactive user support**
+
+            If `user` is authenticated but inactive at the same time, all checks
+            always return `False`.
+
+            Note:
+               Remember, that if user is not *active*, all checks would return `False`.
+
+            Parameters:
+                user_obj (User): User instance.
+                perm (str): Permission string.
+                obj (Model): Model instance.
+
+            Returns:
+                `True` if `user_obj` has permission, `False` otherwise.
+            """
+            from asgiref.sync import sync_to_async
+
+            return await sync_to_async(self.has_perm)(user_obj, perm, obj)
+
     def get_user_permissions(self, user_obj: Any, obj: Optional[Model] = None) -> Iterable[str]:
         """Returns user permissions for a given object.
 
@@ -161,6 +195,7 @@ class ObjectPermissionBackend(BaseBackend):
         return set(check.get_group_perms(obj))
 
     def get_all_permissions(self, user_obj: Any, obj: Optional[Model] = None) -> Iterable[str]:
+        # TODO: since we inherit from BaseBackend, this method should not be necessary
         """Returns all permissions for a given object.
 
         Parameters:
--- a/guardian/backends.py
+++ b/guardian/backends.py
@@ -110,6 +110,32 @@ class ObjectPermissionBackend(BaseBackend):
         check = ObjectPermissionChecker(user_obj)
         return check.has_perm(perm, obj)
 
+    def get_user_permissions(self, user_obj: Any, obj: Optional[Model] = None) -> Iterable[str]:
+        """Returns user permissions for a given object.
+
+        Parameters:
+            user_obj (User): User instance.
+            obj (Model): Django Model instance. If None, returns empty set
+                        since this backend only handles object-level permissions.
+
+        Returns:
+             a set of permission strings that the given specific `user_obj` has for `obj`.
+        """
+        # This backend only handles object-level permissions
+        if obj is None:
+            return set()
+
+        # check if user_obj and object are supported
+        support, user_obj = check_support(user_obj, obj)
+        if not support:
+            return set()
+
+        check = ObjectPermissionChecker(user_obj)
+        if user_obj.is_superuser:
+            return set(check.get_perms(obj))
+
+        return set(check.get_user_perms(obj))
+
     def get_group_permissions(self, user_obj: Any, obj: Optional[Model] = None) -> Iterable[str]:
         """Returns group permissions for a given object.
 
--- a/guardian/backends.py
+++ b/guardian/backends.py
@@ -63,6 +63,15 @@ class ObjectPermissionBackend(BaseBackend):
     supports_anonymous_user = True
     supports_inactive_user = True
 
+    def __getattribute__(self, name) -> Any:
+        if name in ("get_user", "aget_user"):
+            # Django's test Client.force_login and AsyncClient.aforce_login authenticate
+            # using the first backend that implements get_user or aget_user. Since this
+            # backend should not be responsible for authenticating users, we hide those
+            # methods so Django falls back to a backend that does.
+            raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
+        return super().__getattribute__(name)
+
     def has_perm(self, user_obj: Any, perm: str, obj: Optional[Model] = None) -> bool:
         """Check if a user has the permission for a given object.
 
--- a/guardian/backends.py
+++ b/guardian/backends.py
@@ -1,3 +1,4 @@
+from functools import wraps
 from typing import Any, Iterable, Optional
 
 from asgiref.sync import sync_to_async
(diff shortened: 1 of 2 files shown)


Natgho:
Hi @hartungstenio, thank you for your valuable contribution. Could you please write your tests as well?

Natgho:
Hi @hartungstenio , is there any update? (=

hartungstenio:
Hi @Natgho

I was busy working on object permissions using async django-guardian in the company I work to, by inheriting from ObjectPermissionBackend and adding the required methods. It worked.

Will try to bring it to `django-guardian` this week.

guardian/backends.py
```diff
@@ -113,6 +116,63 @@ def has_perm(self, user_obj: Any, perm: str, obj: Optional[Model] = None) -> boo
         check = ObjectPermissionChecker(user_obj)
         return check.has_perm(perm, obj)
 
+    if _supports_async:
+
+        async def ahas_perm(self, user_obj: Any, perm: str, obj: Optional[Model] = None) -> bool:
```
hartungstenio: The default implementation checks if `perm in await self.get_all_permissions(...)`. Keeping it would be a problem since we currently do not return the full permission name (including the `app_label`) on those methods. 🤔

guardian/testapp/tests/test_backends.py
```diff
@@ -44,20 +59,157 @@ def test_has_perm_with_object(self):
         assign_perm("change_project", self.user, self.project)
         self.assertTrue(self.backend.has_perm(self.user, "change_project", self.project))
 
+    @skipIf(django.VERSION < (5, 2), "Async backends are supported on Django 5.2 and above")
+    async def test_ahas_perm_with_object(self):
```
hartungstenio, line 86: I've created an `async` version of each test. There is a lot of duplication, but they are here in case we want to implement native async support (without recurring to `sync_to_async`) someday.

guardian/backends.py
```diff
@@ -113,6 +111,65 @@ def has_perm(self, user_obj: Any, perm: str, obj: Optional[Model] = None) -> boo
         check = ObjectPermissionChecker(user_obj)
         return check.has_perm(perm, obj)
 
+    if django.VERSION >= (5, 2):
```
dpgraham4401: Hey @hartungstenio why do we need this? I'm not at my computer, but wouldn't `ahas_perm` only ever be called from other async backend methods which were only added in 5.2? Or by a downstream user who's using >= 5.2 features.

guardian/backends.py
```diff
@@ -113,6 +111,65 @@ def has_perm(self, user_obj: Any, perm: str, obj: Optional[Model] = None) -> boo
...
+                        since this backend only handles object-level permissions.
+
+        Returns:
+             a set of permission strings that the given specific `user_obj` has for `obj`.
+        """
+        # This backend only handles object-level permissions
+        if obj is None:
+            return set()
+
+        # check if user_obj and object are supported
```
dpgraham4401: Thank you for the comments, but i think we got that the `check_support` function is checking support 😁

guardian/backends.py
```diff
@@ -138,6 +195,7 @@ def get_group_permissions(self, user_obj: Any, obj: Optional[Model] = None) -> I
         return set(check.get_group_perms(obj))
 
     def get_all_permissions(self, user_obj: Any, obj: Optional[Model] = None) -> Iterable[str]:
+        # TODO: since we inherit from BaseBackend, this method should not be necessary
```
dpgraham4401, line 210: I'd have to look again when at a computer. We're also checking if the object supports object level permissions so we can't get rid of it outright, but we should be able to call the super class's method and reduce some boilerplate.

guardian/backends.py
```diff
@@ -113,6 +111,65 @@ def has_perm(self, user_obj: Any, perm: str, obj: Optional[Model] = None) -> boo
...
+
+            Parameters:
+                user_obj (User): User instance.
+                perm (str): Permission string.
+                obj (Model): Model instance.
+
+            Returns:
+                `True` if `user_obj` has permission, `False` otherwise.
+            """
+            from asgiref.sync import sync_to_async
```
dpgraham4401: Why import this inside the function instead of the import block?

dpgraham4401 (changes requested):
Hey @hartungstenio thanks for doing this, I can't believe we weren't already inheriting from BaseBackend. 

I've left some initial comments but don't feel like you need to address them ASAP. I need double check some stuff when I'm back at a computer next week and I don't want to make you feel like we're going back and forth on this PR.

Cheers!

guardian/backends.py
hartungstenio: We don't actually need this. I just wanted to "play safe" with previous versions of Django, but there is no harm in leaving it available to every version.

guardian/backends.py
hartungstenio: Because of that `if django.VERSION >= (5, 2):` a few lines above. After removing that check, I can move this import to module level.

guardian/backends.py
hartungstenio: The super class basically returns a set of the results `get_user_permissions` and `get_group_permissions`, and both call `check_support`.

`ObjectPermissionChecker.get_perms` seems to be doing the same thing, concatenating both user and group permissions.

dpgraham4401: Yeah, since we're using `check_support` and the `ObjectPermissionChecker` in both the `get_user_permissions` and `get_group_permissions`, let get rid of this method and let `BaseBackend` handle that. 👍

dpgraham4401:
@hartungstenio since we're inheriting from `BaseBackend`, looks like we can take advantage of the parent class's implementation. The only 2 items to consider is how much of the backend we want to document (documenting methods implemented by the parent class?), and since this is a core class that we need to make sure is well tested before releasing.

sevdog:
Please look also at #771 and the reason for which I closed that PR.

Natgho:
> Please look also at #771 and the reason for which I closed that PR.

This is a rather critical observation. @hartungstenio Are you sure you want to continue improving this PR? Because it seems like the danger @sevdog mentioned is a deal-breaker.

hartungstenio:
I reported this issue to django, but they closed it 😢. I will add it to the new-features board as they suggested.
https://code.djangoproject.com/ticket/36837

I could change this PR to implement async versions of every function, without inheriting from BaseBackend, but maybe this is part of a larger plan about how to support django async.

What I usually see in django itself is to start by adding async versions simply by calling the regular method through `sync_to_async`, them slowly implementing true async support by actually using the async ORM interface, without `sync_to_async` (at least in django-guardian, django itself still uses sync_to_async on the ORM). The issue with this approach is that it introduces a lot of code duplication (the same logic, but changing a `for` by an `async for` somewhere).

guardian/backends.py
```diff
@@ -55,15 +56,21 @@ def check_support(user_obj: Any, obj: Model) -> Any:
...
-    def authenticate(self, request: HttpRequest, username: Optional[str] = None, password: Optional[str] = None) -> Any:
-        return None
+    def __getattribute__(self, name) -> Any:
+        if name in ("get_user", "aget_user"):
+            # Django's test Client.force_login and AsyncClient.aforce_login authenticate
+            # using the first backend that implements get_user or aget_user. Since this
+            # backend should not be responsible for authenticating users, we hide those
+            # methods so Django falls back to a backend that does.
+            raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
+        return super().__getattribute__(name)
```
hartungstenio: @Natgho @sevdog What do you think of this workaround? This would allow us to inherit from `BaseBackend` (thus getting async support) while still working with `(a)force_login`

Natgho:
@hartungstenio I appreciate the effort and dedication you’ve put into this PR. 

I’ve responded to the Django team in the ticket. I’ve also reviewed this ticket for you using AI. Please implement the changes the AI mentioned. Additionally, please open the new discussion ticket mentioned by the Django team, include the comment I wrote, and help facilitate the Django team’s action on this matter. Thanks to their support, we will be able to integrate this feature into the django-guardian side as well. 

Please review the “conversations” initiated by the AI and proceed iteratively with the Agent until it is finalized.

Natgho:
https://github.com/django/new-features/issues/130

guardian/backends.py
sevdog: This could work, yet IMO it is a bad way to do such thing: `__getattribute__` will be called for every other attribute on the class, slowing it down.

A "Faster" approach would be to fool the `hasattr` method to belive that these method are not implemented like this:

```suggestion
    @property
    def get_user(self) -> None:
        raise AttributeError(f"'{type(self).__name__}' object has no attribute 'get_user'")

    @property
    def aget_user(self) -> None:
        raise AttributeError(f"'{type(self).__name__}' object has no attribute 'aget_user'")
```

Because [`hasattr`](https://docs.python.org/3/library/functions.html#hasattr) actually calls the "attribute" to see it raises an `AttributeError`.

guardian/backends.py
```diff
@@ -45,6 +47,14 @@ def check_user_support(user_obj: Any) -> tuple[bool, Any]:
     return True, user_obj
 
 
+def noop(f: Any) -> Any:
+    @wraps(f)
+    def raise_attr_error(self, *args: Any, **kwargs: Any):
+        raise AttributeError(f"'{type(self).__name__}' object has no attribute '{f.__name__}'")
+
+    return property(raise_attr_error)
```
hartungstenio, line 55: @Natgho I was thinking we could use a decorator to make it easier to add .noop = True support later once Django supports it.

```python
def noop(f: Any) -> Any:
    if django.VERSION < (6, 1):
        @wraps(f)
        def raise_attr_error(self, *args: Any, **kwargs: Any):
            raise AttributeError(
                f"'{type(self).__name__}' object has no attribute '{f.__name__}'"
            )
        return property(raise_attr_error)
    else:
        f.noop = True
        return f
```

This keeps compatibility with older Django versions while transparently switching to the native `.noop` opt-out when available — no changes needed at the call sites.

Natgho: Good idea. That makes perfect sense.

jacobtylerwalls: Hey folks, I'm sprucing up https://github.com/django/django/pull/21065, and based on the feedback there, we're going to do this with an identity check (`is`) instead of a `noop` attribute.

guardian/backends.py
```diff
@@ -55,15 +65,20 @@ def check_support(user_obj: Any, obj: Model) -> Any:
...
 
-    def authenticate(self, request: HttpRequest, username: Optional[str] = None, password: Optional[str] = None) -> Any:
-        return None
+    @noop
+    def get_user(self, user_id: int) -> None:
+        pass
+
+    @noop
+    async def aget_user(self, user_id: int) -> None:
+        pass
```
jacobtylerwalls: With the latest revisions to django/django#21065, this would need to be deleted. Would that still work for you?
```suggestion
```

sevdog: Since that PR may be released in django 6.1 (or at minimum in 6.0) and guardian still has support for django 3.2, it would be nice to put an IF to define this _noop_ methods only when installed django version does not support that feature.

hartungstenio: @sevdog Done. It will be released on Django 6.2

Natgho:
Hi @hartungstenio, first of all, thank you for the contribution...

I know this PR has been sitting around for a while. Every time I try to read through and test this PR, I find myself working on something else entirely. 

That’s why I need some help. If at least one more person who’s available could test and verify it, I’d be much more eager to accept a PR like this.

@dpgraham4401 @sevdog @thclark If anyone is available, I’d appreciate it if they could check it.

Natgho:
Hi @hartungstenio , Please take the following suggestions into consideration;

## `guardian/backends.py`

### Lines around the `noop` helper
**What to change**
- Rename `noop` to something descriptive, such as `hide_attribute` or `raise_attribute_error`.
- Add a short docstring explaining that it deliberately raises `AttributeError` so `hasattr()` returns `False`.

**Why**
- The current name is misleading: this function is not a no-op.
- Future maintainers may “simplify” it and unintentionally break the backend-selection behavior.

**Suggested replacement**
```python
def hide_attribute(f: Any) -> Any:
    """Hide an attribute by making hasattr() return False."""
    @wraps(f)
    def raise_attr_error(self, *args: Any, **kwargs: Any):
        raise AttributeError(f"'{type(self).__name__}' object has no attribute '{f.__name__}'")

    return property(raise_attr_error)
```

### Lines around the Django version check for `get_user` / `aget_user`
**What to change**
- Replace tuple comparisons that depend on `django.VERSION` length with an explicit two-part version check, for example `django.VERSION[:2] < (6, 2)`.

**Why**
- Direct tuple comparisons against `django.VERSION` can behave unexpectedly on patch releases.
- Using `[:2]` makes the intent explicit and prevents accidental skipping or enabling on the wrong Django patch version.

**Suggested replacement**
```python
if django.VERSION[:2] < (6, 2):
```

### Lines around `get_user` and `aget_user`
**What to change**
- Re-evaluate whether these methods need to be hidden at all.
- If the goal is only to rely on `BaseBackend` defaults, prefer the inherited implementation instead of masking the methods with an `AttributeError`-raising property.
- If the workaround is truly required, add a clear inline comment explaining the Django auth behavior being worked around.

**Why**
- This pattern is fragile and not obvious.
- It is easy for someone to misread this as dead code and remove it.
- If Django’s `BaseBackend` already provides the needed behavior, the workaround may be unnecessary complexity.

**Suggested replacement**
```python
# Django backend discovery may treat get_user/aget_user differently, so we
# intentionally hide these methods to avoid this backend being used for login flows.
```

### Lines around `ahas_perm`
**What to change**
- Consider whether the explicit `sync_to_async(self.has_perm)` wrapper is needed.
- If Django’s `BaseBackend` async support already provides the correct behavior, rely on that instead of adding another wrapper.

**Why**
- A manual async wrapper may duplicate Django’s built-in async backend support.
- Reducing custom async plumbing lowers the chance of subtle threading or connection issues.

**Suggested replacement**
```python
async def ahas_perm(self, user_obj: Any, perm: str, obj: Optional[Model] = None) -> bool:
    """Async version of has_perm using sync_to_async for compatibility."""
    return await sync_to_async(self.has_perm)(user_obj, perm, obj)
```

If you keep the current approach, the docstring should clearly say why it exists.

### Lines around `get_user_permissions`
**What to change**
- Add a short comment or test clarifying the superuser branch.
- Make sure the behavior is documented: superusers use `check.get_perms(obj)` while regular users use `check.get_user_perms(obj)`.

**Why**
- The logic is valid, but subtle.
- Without explanation, this branch is easy to misinterpret later.

**Suggested replacement**
```python
# Superusers get all object permissions for the target object.
if user_obj.is_superuser:
    return set(check.get_perms(obj))
```

### Lines around the `get_all_permissions` TODO
**What to change**
- Remove the TODO or replace it with a precise explanation.
- If the method is still required for compatibility, explain that explicitly.

**Why**
- The current TODO is vague and not actionable.
- It suggests uncertainty about the method’s purpose.

**Suggested replacement**
```python
# This backend only handles object-level permissions, but we keep this override
# to match Django's backend API and preserve compatibility.
```

---

## `guardian/testapp/tests/test_backends.py`

### Lines around `test_get_user_not_exists`
**What to change**
- Avoid testing `hasattr()` directly if possible.
- Prefer a behavior-based test that checks the backend’s observable auth behavior instead of its internal attribute visibility.

**Why**
- This test currently checks an implementation detail.
- It may become brittle if Django changes how it inspects auth backends.

**Suggested replacement**
```python
def test_get_user_is_not_used_for_auth_flow(self):
    """Backend should not participate in user retrieval/auth flow."""
    self.assertFalse(hasattr(self.backend, "get_user"))
```

If a stronger behavior-based test is available, use that instead.

### Lines around `test_aget_user_not_exists`
**What to change**
- Same as the sync version: avoid relying only on `hasattr()`.
- Prefer a test of actual backend behavior if possible.

**Why**
- Implementation-detail tests are easier to break accidentally and harder to interpret.

### Lines around the Django `skipIf` / `skipUnless` decorators
**What to change**
- Replace version guards that compare `django.VERSION` directly with guards that use `django.VERSION[:2]`.
- If you need both a lower and upper bound, use explicit two-part comparisons.

**Why**
- The current comparisons may skip patch releases unintentionally.
- Explicit major/minor comparisons are clearer and safer.

**Suggested replacement**
```python
@skipUnless((5, 2) <= django.VERSION[:2] < (6, 2), "Django 5.2+ async backend support")
```

### Lines around `test_ahas_perm_with_object`
**What to change**
- Add a parity assertion that compares sync and async results for the same setup.

**Why**
- This strengthens confidence that `ahas_perm()` behaves exactly like `has_perm()`.

**Suggested replacement**
```python
self.assertEqual(
    self.backend.has_perm(self.user, "change_project", self.project),
    await self.backend.ahas_perm(self.user, "change_project", self.project),
)
```

### Lines around `test_get_user_permissions_superuser`
**What to change**
- Replace the “non-empty” assertion with a concrete expected permission assertion if possible.

**Why**
- A non-empty set is a weak guarantee.
- A specific expected permission makes the test more deterministic.

**Suggested replacement**
```python
self.assertIn("change_project", perms)
```

### Lines around `test_get_user_permissions_without_object`
**What to change**
- Ensure the test is actually calling `get_user_permissions(self.user)` and not another permissions method.

**Why**
- This test should validate the user-permissions branch directly.
- Calling the wrong method would leave the intended behavior untested.

**Suggested replacement**
```python
perms = self.backend.get_user_permissions(self.user)
self.assertEqual(set(), perms)
```

### Lines around `test_aget_user_permissions_without_object`
**What to change**
- Ensure the async version also calls `aget_user_permissions(self.user)` directly.

**Why**
- The async user-permissions path should be covered explicitly.
- Otherwise the test does not validate the intended code path.

**Suggested replacement**
```python
perms = await self.backend.aget_user_permissions(self.user)
self.assertEqual(set(), perms)
```

### Lines around the async mixed-permissions and removal tests
**What to change**
- Add a small parity check showing that sync and async permission results match for the same setup.

**Why**
- This makes it easier to trust the new async backend behavior.
- It reduces the risk of async and sync logic drifting apart.

**Suggested replacement**
```python
sync_perms = self.backend.get_all_permissions(self.user, self.project)
async_perms = await self.backend.aget_all_permissions(self.user, self.project)
self.assertEqual(sync_perms, async_perms)
```

---

## Overall recommendation
Before merging, I would ask for:
1. The Django version guards to be corrected.
2. The `get_user` / `aget_user` hiding workaround to be clarified or simplified.
3. The TODO in `get_all_permissions` to be made concrete.
4. A couple of tests to be rewritten to focus on behavior rather than implementation detail.

PR #953, opened 2026-01-13
make shortcuts.assign_perm idempotent when doing bulk operations

nigelzor:
matching the behaviour observed when assigning them one-by-one

previously, passing a list or queryset into shortcuts.assign_perm could throw if the user or group already had the permission granted.


Natgho (commented):
LGTM

PR #954, opened 2026-01-13
save one database query per assign_perm when using generic permissions

nigelzor:
`get_or_create` with a generic foreign key will try to look up the content_object by id if it's not provided. we already have it, so let's avoid that query.


Natgho:
Hi @nigelzor , Please fix the parts that are causing errors in the tests. Thank you for your contribution 💯

nigelzor:
I've updated the test to use `ContentType` instead of `Project` so that it's clearer that it's testing `UserObjectPermission` and not `ProjectUserObjectPermission`.

Natgho (approved):
LGTM

PR #956, opened 2026-02-24
DO NOT MERGE - Updates to inactive user filtering

thclark:
## TLDR

This is an unholy beast of a PR. Please let us not merge this.

## Summary

This PR builds on @Natgho 's contribution in #952, which in turn builds on the bug noted by @raphaelmerx in #854.

[Here](https://github.com/django-guardian/django-guardian/pull/854#discussion_r1986917346), @BonaFideIT suggests some sensible basic requirements:
- introduce a settings that allows activation of the behaviour
- add a check if the user model has the property "is_active"
- add a warning if the setting is used and is_active is not present.

So, I did a deep dive into consistency with results reported [here](https://github.com/django-guardian/django-guardian/pull/952#pullrequestreview-3848826619). Between us, @Natgho and I have implemented the above, whilst making every permission check consistent in behaviour with respect to `user.is_active`.

Unfortunately, **either** we accept inconsistency in behaviour across the various checks, **or** we need a breaking change.

### ⚠️ 💥 Breaking change is required 

Previously, `has_perm()`, `get_perms()`, etc. **unconditionally** returned empty results for inactive users, whilst other check tools like the shortcuts **did not**, as @raphaelmerx pointed out.

In this PR, with the newly added setting `GUARDIAN_ACTIVE_USERS_ONLY=False` (the default), inactive users are no longer blocked by ObjectPermissionChecker. Now, that behavior only happens when `GUARDIAN_ACTIVE_USERS_ONLY=True`. The entire inactive-user filtering story is now controlled by a single setting, consistently across all APIs.

The alternative was to set `GUARDIAN_ACTIVE_USERS_ONLY=True` by default, which maintains compatibility on the `ObjectPermissionChecker` level but breaks it on the shortcuts in exactly the same way as the original PR.

Whilst this works, there is a hell of a lot of code here, a ton of edge cases, a lot of extra logic... __*all for a feature nobody asked for*__ - even the originator just wanted consistent behaviour with `has_perm`, they didn't need to control the `is_active` behaviour -- that's just something we cooked up to avoid making a breaking change.

### Other notes

- guardian already presupposes `is_active` on a user model ([f.ex here](https://github.com/django-guardian/django-guardian/blob/77c7a8b666321900309212d73c36d68fb5789aef/guardian/core.py#L76)) so it's not actually breaking to continue using it (although in this PR for the sake of form, I checked for its presence).

## API Consistency and Suggested Alternative

It's taken five hours to build this plus whatever time @Natgho spent... I really don't want to spend  the next five years maintaining it. 

**Long term, I don't believe Guardian will be well served by inconsistent behaviour in the API. so I don't think that we should accept inconsistency to avoid a breaking change.**

So, given that we have to make a breaking change anyway I recommend that we basically implement Raphael's suggestion in all cases (inactive users never given permission, not adjustable), ensure that's part of all the shortcuts uniformly... then release that as part of Version 4.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -281,3 +281,61 @@ GroupObjectPermission = get_user_obj_perms_model()
 ```
 
 Defaults to `'guardian.GroupObjectPermission'`.
+
+## `GUARDIAN_WORK_ONLY_ACTIVE_USERS`
+
+!!! abstract "Added in version 2.4.0"
+
+When set to `True`, functions that return users with permissions (like
+`get_users_with_perms`) will only return active users (`is_active=True`).
+This filter applies to all permission-related functions that retrieve users,
+including direct user permissions, group permissions, and superuser permissions.
+
+This setting is particularly useful in applications where inactive users should
+not be considered when retrieving users with permissions, even if they technically
+still have permission assignments in the database.
+
+```python
+# Default behavior - return all users with permissions
+GUARDIAN_WORK_ONLY_ACTIVE_USERS = False
+
+# Only return active users with permissions
+GUARDIAN_WORK_ONLY_ACTIVE_USERS = True
+```
+
+!!! example "Usage example"
+
+    ```python
+    from guardian.shortcuts import get_users_with_perms, assign_perm
+    from django.contrib.auth import get_user_model
+
+    User = get_user_model()
+
+    # Create users
+    active_user = User.objects.create_user('active', is_active=True)
+    inactive_user = User.objects.create_user('inactive', is_active=False)
+
+    # Assign permissions to both users
+    assign_perm('change_model', active_user, my_object)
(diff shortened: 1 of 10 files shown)


Natgho:
Tom... That's an incredible effort, congratulations, but you're right. Sometimes stop-loss is good. Let's put this on hold for now. I agree that the inconsistency you identified is unacceptable and cannot be overlooked in security software. If necessary, let's reduce the PRs to a single PR and keep it as a reference point for anyone who wants to deal with this madness in the future and has the time.

dpgraham4401:
This is great y'all! 🚀 

Since we're probably going to have to cut a breaking change release anyway after this project finds it's way into main, would it make sense to break towards the secure by default appraoch and use `GUARDIAN_ACTIVE_USERS_ONLY=True` as the default?

PR #995, opened 2026-06-08
Fix attached permissions when groups are excluded

marko1olo:
## Summary
- respect `with_group_users=False` when `attach_perms=True` and `with_superusers=True` are used together
- keep full permission lists for included superusers
- add regression coverage for a direct-permission user who also belongs to a permitted group
- add the configured git identity to `pyproject.toml` per the contribution guide

## Tests
- `.\.venv\Scripts\python.exe -m pytest guardian\testapp\tests\test_shortcuts.py::GetUsersWithPermsTest::test_without_group_users_with_superusers_and_perms_attached guardian\testapp\tests\test_shortcuts.py::GetUsersWithPermsTest::test_direct_perms_only_perms_attached guardian\testapp\tests\test_shortcuts.py::GetUsersWithPermsTest::test_without_group_users_no_result_but_with_superusers -q`
- `.\.venv\Scripts\python.exe -m pytest guardian\testapp\tests\test_shortcuts.py -q`
- `.\.venv\Scripts\python.exe -m pytest guardian\testapp\tests\test_direct_rel.py -q`
- `.\.venv\Scripts\ruff.exe check guardian\shortcuts.py guardian\testapp\tests\test_shortcuts.py`
- `.\.venv\Scripts\ruff.exe format --check guardian\shortcuts.py guardian\testapp\tests\test_shortcuts.py`
- `PYTHONPYCACHEPREFIX=C:\tmp\pycache-django-guardian-superuser-perms .\.venv\Scripts\python.exe -m compileall -q guardian\shortcuts.py guardian\testapp\tests\test_shortcuts.py`
- `git diff --check`

Notes: pytest/ruff cache writes emitted local Windows access-denied cache warnings in this sandbox, but the test and lint assertions above passed.

### DIFF
--- a/guardian/shortcuts.py
+++ b/guardian/shortcuts.py
@@ -435,8 +435,7 @@ def get_users_with_perms(
             only_with_perms_in=only_with_perms_in,
             with_superusers=with_superusers,
         ):
-            # TODO: Support the case of set with_group_users but not with_superusers.
-            if with_group_users or with_superusers:
+            if with_group_users or (with_superusers and user.is_superuser):
                 users[user] = sorted(get_perms(user, obj))
             else:
                 users[user] = sorted(get_user_perms(user, obj))
--- a/guardian/testapp/tests/test_shortcuts.py
+++ b/guardian/testapp/tests/test_shortcuts.py
@@ -854,6 +854,28 @@ class GetUsersWithPermsTest(TestCase):
         expected = {self.user1: ["change_contenttype"], admin: ["delete_contenttype"]}
         self.assertEqual(result, expected)
 
+    def test_without_group_users_with_superusers_and_perms_attached(self):
+        admin = User.objects.create(username="admin", is_superuser=True)
+        self.user1.groups.add(self.group1)
+        self.user2.groups.add(self.group1)
+        assign_perm("change_contenttype", self.user1, self.obj1)
+        assign_perm("delete_contenttype", self.group1, self.obj1)
+
+        result = get_users_with_perms(self.obj1, attach_perms=True, with_superusers=True, with_group_users=False)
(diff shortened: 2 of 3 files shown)


marko1olo:
I checked the ReadTheDocs failure on this PR. The raw log for build `33036431` fails during `mkdocs build` after the privacy plugin times out retrieving the GitHub Actions badge:

```text
WARNING - Couldn't retrieve https://github.com/django-guardian/django-guardian/workflows/Tests/badge.svg?branch=devel: Read timed out
FileNotFoundError: .../.cache/plugin/privacy/assets/external/github.com/django-guardian/django-guardian/workflows/Tests/badge.16c1ad22.svg
```

That path is from the docs external-asset cache and is unrelated to the `guardian.shortcuts.get_users_with_perms()` change in this PR.

Natgho:
Thank you for the development, but this behavior change is quite risky for a project like DG that has been in use for many years.

Some projects may have unknowingly become dependent on the current “buggy” behavior. For example, code that expects group permissions to be returned for regular users even though `with_group_users=False` is set
In such projects, the result set may change. also using configurations like the following:
```
attach_perms=True
with_superusers=True
with_group_users=False
```
the permission lists returned may now be narrower. While this is the correct behavior, it could break snapshots, tests, or assertions.

On the other hand, there may be unexpected side effects in other combinations of the function (only_with_perms_in, anonymous users, global permissions, etc.), so the tests may also need to be improved.

Taking all this into account, merging this PR involves excessive risk, so we cannot release it.

marko1olo:
Thanks for the detailed review. That backward-compat risk makes sense for this API surface. I will leave this closed rather than trying to push the behavior change further.

PR #994, opened 2026-06-05
Fix permission filtering for joined querysets

marko1olo:
## Summary

Fix `filter_perms_queryset_by_objects` so object querysets with existing joins still produce a single-column subquery for `__in` lookups.

The generic-object branch previously used `.values(_pk=...)`, and joined querysets can keep extra selected columns in that subquery, causing:

```text
ValueError: The QuerySet value for the 'in' lookup must have 1 selected fields
```

This changes the object primary-key subqueries to use `.values_list(..., flat=True)` after annotating the transformed primary key.

There is a nearby open PR (#990) for the bigint/object_pk cast regression, but it does not cover this joined-queryset failure mode and still uses `.values(...)` in this function.

## Test plan

- `.\.venv\Scripts\python.exe -m pytest guardian\testapp\tests\test_shortcuts.py::GetObjectsForUser::test_joined_queryset_with_cast_primary_key guardian\testapp\tests\test_shortcuts.py::GetObjectsForUser::test_filter_perms_queryset_by_joined_objects guardian\testapp\tests\test_shortcuts.py::GetObjectsForUser::test_uuid_primary_key guardian\testapp\tests\test_shortcuts.py::GetObjectsForUser::test_varchar_primary_key -q` -> `4 passed`
- `python -m ruff check guardian\shortcuts.py guardian\testapp\tests\test_shortcuts.py`
- `python -m ruff format --check guardian\shortcuts.py guardian\testapp\tests\test_shortcuts.py`
- `git diff --cached --check` before commit

Fixes #992.

PR #1008, opened 2026-08-14
feat: inherit ObjectPermissionBackend from Django BaseBackend

### LINKED ISSUE #947: Make ObjectPermissionBackend inherit from BaseBackend

marko1olo:
## Description

Fixes #947.

Makes `ObjectPermissionBackend` inherit from `django.contrib.auth.backends.BaseBackend`.

## Motivation

In modern Django, standard authentication and authorization backends inherit from `BaseBackend`, providing default hook implementations (`get_user`, `has_perm`, `has_module_perms`, `get_all_permissions`, etc.) and satisfying `isinstance(backend, BaseBackend)` checks across third-party auth tooling and Django internals.

## Verification

- Added unit test `test_inherits_base_backend` in `guardian/testapp/tests/test_backends.py`.
- All 17 tests in `test_backends.py` pass cleanly.


Natgho:
https://github.com/django-guardian/django-guardian/pull/994#issuecomment-5357040509

PR #1010, opened 2026-08-24
Add opt-in ReinforcedGuardedModelAdminMixin

aroh3006:
Related to #756.

GuardedModelAdminMixin only checks has_change_permission on the underlying model for its object permissions management views. Any staff user allowed to change a model instance can view and edit every user's and group's guardian permissions for it, even without the guardian view/change permission for that.

A previous fix for this changed the default behavior and was reverted (#757, #764), since some applications might rely on it and the change wasn't something everyone agreed should be forced on all users of GuardedModelAdminMixin. This adds the same check as an opt-in mixin instead. Existing GuardedModelAdminMixin/GuardedModelAdmin subclasses are unaffected.

ReinforcedGuardedModelAdminMixin and ReinforcedGuardedModelAdmin wrap each of the three object permission views with the matching guardian permission check before delegating to the base implementation. Added tests covering both the denied and allowed case for all three views, plus a docs section and API reference entries.

Ran the full admin test suite (56 passed) and ruff. I also temporarily removed the permission checks and confirmed the new tests then fail (302 expected, got 200). That confirms they actually exercise the fix rather than just checking the class exists.

### DIFF
--- a/docs/api/admin.md
+++ b/docs/api/admin.md
@@ -3,3 +3,7 @@
 ::: guardian.admin.GuardedModelAdmin
 
 ::: guardian.admin.GuardedModelAdminMixin
+
+::: guardian.admin.ReinforcedGuardedModelAdmin
+
+::: guardian.admin.ReinforcedGuardedModelAdminMixin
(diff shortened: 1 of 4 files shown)


Natgho:
Hi @aroh3006 , thank you for your contribution. Please make the correction suggested by the AI. Then let's review it again.

Natgho (approved):

Natgho:
Thank you for your valuable contribution 🙏

aroh3006:
Really glad I could contribute @Natgho

PR #1019, opened 2026-09-01
Collapse integer PK field tuple in _handle_pk_field

sevdog:
`AutoField`, `BigIntegerField`, `PositiveIntegerField`,`PositiveSmallIntegerField` and `SmallIntegerField` **all** subclass `IntegerField`, so the tuple check was fully redundant.


Natgho:
Maybe a few tests could be added while this change is being made?

sevdog:
There were already no tests to ensure the coverage for the specific integer types, thus the branch was not well covered from the beginning.

I dunno which value could be added by a specific test here since django already tests type inheritance for these fields.

Natgho (approved):

